PR #1, opened 2023-03-02
Scaffolding

saraswatpuneet:
- [x]  Basic action
- [x] Make file
- [x] toolchain config file
- [x] fmt config file
- [x] placeholder core library
- [x] Added branch rules and protection rules for main
- [x] License, Contributing, tool-versions

Anything else?


.tool-versions
```diff
@@ -0,0 +1,6 @@
+rust 1.64
+make 4.3
+cmake 3.24.1
```
aramikm, line 3: do we need `cmake` and `protoc` even if when we are not using substrate?

core/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "graph-sdk-core"
+edition = "2021"
+homepage = "https://spec.dsnp.org"
+license = "Apache-2.0"
+publish = false
+repository = "https://github.com/libertyDSNP/graph-sdk/"
+version = "0.0.0"
+
+[dependencies]
+wasm-bindgen = {version = "0.2.63", optional = true}
```
aramikm, line 11: do we want to include wasm dependencies right now or just wait and add whenever we wanted to add support for wasm?

aramikm (commented):
Added some questions and I think we might want to add some of the following also:
- LICENSE
- .gitignore

aramikm (approved):
Looks good!

core/Cargo.toml
saraswatpuneet: I would say keep it, ideally the rust code should be wasm  compatible and decorated accordingly
can come in later too, so no hard preferences

.tool-versions
saraswatpuneet: cmake maybe , protoc no, unless any dependency we may have needs it

saraswatpuneet:
@aramikm  Readme, .gitingore and License files are already set in main

PR #2, opened 2023-03-16
Init schemas

aramikm:
# What
Creating a utility to read and write from and using avro schemas.
closes https://github.com/LibertyDSNP/frequency/issues/1144


core/src/dsnp/schema.rs
```diff
@@ -0,0 +1,181 @@
...
+		Schema::parse_str(include_str!("../../../schemas/user_public_graph_chunk.json"))
+			.expect("user_public_graph_chunk.json is not avro compatible");
+	static ref PUBLIC_GRAPH_SCHEMA: Schema =
+		Schema::parse_str(include_str!("../../../schemas/public_graph.json"))
+			.expect("public_graph.json is not avro compatible");
+	static ref PRIVATE_GRAPH_CHUNK_SCHEMA: Schema =
+		Schema::parse_str(include_str!("../../../schemas/user_private_graph_chunk.json"))
+			.expect("user_private_graph_chunk.json is not avro compatible");
+}
+
```
saraswatpuneet, line 25: you may want to clean it up like this for example

https://github.com/LibertyDSNP/frequency/blob/22aee5f04bfc9ff90a46dc4b9fe5ddcb493c7950/node/service/src/chain_spec/frequency.rs#L112

aramikm: But you are reading and parsing the json anytime `load_frequency_spec()` is called. here we just read and parse it 1 time and can reuse the schema as many times as you want after that.

aramikm: I might be wrong though. let me know what you think

core/src/dsnp/dsnp_types.rs
```diff
@@ -0,0 +1,168 @@
...
+
+/// Prid len in bytes
+const PRID_LEN_IN_BYTES: usize = 8;
+/// Inner Graph type used in both private and public graphs
+pub type DsnpInnerGraph = Vec<DsnpGraphEdge>;
+/// `Pseudonymous Relationship Identifier` which allows private connection verification
+/// Wrapping in its own Struct allows easier serialization and deserialization
+#[derive(Debug, Clone, PartialEq)]
+pub struct DsnpPrid {
+	inner: Vec<u8>,
```
JoeCap08055, line 15: Since we have an upper bound of `PRID_LEN_IN_BYTES`, should we use `BoundedVec<u8, PRID_LEN_IN_BYTES>` here?

JoeCap08055:
I find the `read/write` terminology a little bit confusing. Normally I consider a `read` operation to consume data from some source (optionally mutating the state of the read source, like current pointer, etc), and a `write` operation to send data to some data sink. In the read/write methods here, for `read` we're given a (non-mutable) data buffer to be transformed by decrypting/decompressing/deserializing, and for `write` we're given an object to serialize/compress/encrypt, resulting in a new object/buffer, but not mutating any external source/sink.

Perhaps serialize/deserialize is more appropriate terminology? Or am I not understanding what's happening?

core/src/dsnp/dsnp_types.rs
JoeCap08055: Ah, right, forgot that was not a standard Rust struct.

core/src/dsnp/schema.rs
saraswatpuneet: Aah thats right, it looks good to me

PR #9, opened 2023-03-24
Pri implementation

aramikm:
PRI implementation using Rust native libraries based on following spec definition
https://github.com/LibertyDSNP/spec/blob/main/pages/DSNP/Types/PRId.md


core/src/dsnp/pseudo_relationship_identifier.rs
```diff
@@ -0,0 +1,60 @@
...
+const PRI_CONTEXT: &[u8] = b"PRIdCtx0";
+struct PseudonymousRelationshipIdentifier;
+
+impl PseudonymousRelationshipIdentifier {
+    fn create(a: u64, b: u64, a_secret_key: &SecretKey, b_public_key: &PublicKey) -> DsnpPrid {
+        let id_b = b.to_le_bytes();
+
+        let root_shared = x25519(*a_secret_key.as_bytes(), *b_public_key.as_bytes());
+        let hk = Hkdf::<Sha256>::new(Some(&id_b[..]), &root_shared);
+        let mut derived_key = [0u8; 64];
```
aramikm, line 17: so had to expand the key to 64 since Salsa box needed private and public key. Used the first 32 for public and the next for private. If we use other salsa library we might be able to just use 1 key but didn't want to introduce another library if it was necessary.

core/src/dsnp/pseudo_relationship_identifier.rs
```diff
@@ -0,0 +1,63 @@
...
+const PRI_CONTEXT: &[u8] = b"PRIdCtx0";
+struct PseudonymousRelationshipIdentifier;
+
+impl PseudonymousRelationshipIdentifier {
+	fn create(a: u64, b: u64, a_secret_key: &SecretKey, b_public_key: &PublicKey) -> DsnpPrid {
+		let id_b = b.to_le_bytes();
+
+		let root_shared = x25519(*a_secret_key.as_bytes(), *b_public_key.as_bytes());
+		let hk = Hkdf::<Sha256>::new(Some(&id_b[..]), &root_shared);
+		let mut derived_key = [0u8; 64];
```
aramikm, line 16: so had to expand the key to 64 since Salsa box needed private and public key. Used the first 32 for public and the next for private. If we use other salsa library we might be able to just use 1 key but didn't want to introduce another library if it was necessary.

core/src/dsnp/pseudo_relationship_identifier.rs
```diff
@@ -0,0 +1,63 @@
...
+		hk.expand(PRI_CONTEXT, &mut derived_key).unwrap();
+
+		let mut public_key_arr = [0u8; 32];
+		let mut secret_key_arr = [0u8; 32];
+		public_key_arr.copy_from_slice(&derived_key[..32]);
+		secret_key_arr.copy_from_slice(&derived_key[32..]);
+		let salsa = SalsaBox::new(&public_key_arr.into(), &secret_key_arr.into());
```
aramikm, line 23: can be replaced with `XSalsa20Poly1305::new(&key);` if willing to use that library directly

core/Cargo.toml
```diff
@@ -16,6 +16,15 @@ lazy_static = "1.4.0"
 thiserror = "1.0.39"
 anyhow = "1.0.69"
 miniz_oxide = "0.7.1"
+crypto_box = { version = "0.8.2", features = ["seal"] }
+x25519-dalek = "1"
+hkdf = "0.12.3"
+sha2 = { version = "0.10.6" }
+zeroize = "1.5.7"
+xsalsa20poly1305 = { version = "0.9", features = ["rand_core"], default-features = false}
```
aramikm, line 24: `crypto_box` uses `xsalsa20poly1305` inside so no additional dependency

wesbiggs (approved):

core/Cargo.toml
```diff
@@ -17,6 +17,11 @@ thiserror = "1.0.39"
 anyhow = "1.0.69"
 miniz_oxide = "0.7.1"
 crypto_box = { version = "0.8.2", features = ["seal"] }
+x25519-dalek = "1"
+hkdf = "0.12.3"
+sha2 = { version = "0.10.6" }
+zeroize = "1.5.7"
+xsalsa20poly1305 = { version = "0.9", default-features = false}
```
aramikm, line 24: `crypto_box` uses `xsalsa20poly1305` inside so no additional dependency

core/src/dsnp/pseudo_relationship_identifier.rs
```diff
@@ -0,0 +1,79 @@
...
+	) -> Result<DsnpPrid> {
+		let id_b = b.to_le_bytes();
+
+		// calculate shared secret
+		let root_shared = StaticSecret::from(*a_secret_key.as_bytes())
+			.diffie_hellman(&x25519_dalek::PublicKey::from(*b_public_key.as_bytes()));
+
+		// derive a new key form pri context
+		let mut derived_key = Zeroizing::new([0u8; 32]);
+		let hk = Hkdf::<Sha256>::new(Some(&id_b[..]), root_shared.as_bytes());
```
aramikm, line 29: Unfortunately `Hkdf` does not support `Blake2` for hashing! so had to use `Sha256`

core/src/dsnp/pseudo_relationship_identifier.rs
```diff
@@ -0,0 +1,79 @@
...
+		hk.expand(PRI_CONTEXT, derived_key.as_mut_slice())
+			.map_err(|e| Error::msg(format!("key derivation error {:?}", e)))?;
+
+		// setting nonce with `b` for encryption
+		let mut nonce = [0u8; 24];
+		nonce[16..].copy_from_slice(&id_b[..]);
```
aramikm, line 35: I'm setting the last 8 bytes but I'm not totally sure if this is correct.
@wesbiggs Please take a look.

core/src/dsnp/pseudo_relationship_identifier.rs
```diff
@@ -0,0 +1,79 @@
...
+	fn create(
+		a: u64,
+		b: u64,
+		a_secret_key: &SecretKey,
+		b_public_key: &PublicKey,
+	) -> Result<DsnpPrid> {
+		let id_b = b.to_le_bytes();
+
+		// calculate shared secret
+		let root_shared = StaticSecret::from(*a_secret_key.as_bytes())
```
aramikm, line 24: That extra step that was happening in libsodium was to create a uniform key for encryption but since we are not using that shared key directly and deriving another key via a hash based function. This would make the final key used for encryption a uniform one.

core/src/dsnp/pseudo_relationship_identifier.rs
```diff
@@ -0,0 +1,79 @@
...
+			.diffie_hellman(&x25519_dalek::PublicKey::from(*b_public_key.as_bytes()));
+
+		// derive a new key form pri context
+		let mut derived_key = Zeroizing::new([0u8; 32]);
```
aramikm, line 28: do not need to call `zeroize()` directly since it will be call on `drop()` of each item.

core/src/dsnp/pseudo_relationship_identifier.rs
wesbiggs: Let's use the first 8 bytes (0-7). I will clarify this on the DSNP page.

core/src/dsnp/pseudo_relationship_identifier.rs
wesbiggs: Hmm, that's kind of a problem for compatibility purposes. Is there a workaround?

wesbiggs: Alternatively, we could manually do the libsodium key derivation, which they specify as

```
BLAKE2B-subkeylen(key=key, message={}, salt=subkey_id || {0}, personal=ctx || {0})
```

...probably good to have a test vector to verify against, of course.

(https://doc.libsodium.org/key_derivation)

aramikm: Found a native rust libsodium implementation that seems like it is completely compatible with the C version. So far compatibility testing works as expected
https://crates.io/crates/dryoc

core/src/dsnp/encryption.rs
```diff
@@ -47,9 +58,9 @@ mod tests {
 			83, 98, 0, 10, 234, 88, 23, 54, 23, 23, 109, 198, 111, 70, 2, 89,
 		];
 
-		let secret_key = SecretKey::generate(&mut OsRng);
-		let encrypted = SealBox::encrypt(&plain_data, &secret_key.public_key()).unwrap();
-		let decrypted = SealBox::decrypt(&encrypted, &secret_key).unwrap();
+		let key_pair = StackKeyPair::gen();
+		let encrypted = SealBox::encrypt(&plain_data, &key_pair.public_key).unwrap();
```
JoeCap08055, line 62: Hmm, for unit test shouldn't we call `EncryptionBehavior::encrypt/decrypt` rather than `SealBox` directly?

core/src/dsnp/pseudo_relationship_identifier.rs
```diff
@@ -0,0 +1,91 @@
...
+	keypair::{PublicKey, SecretKey},
+	types::ByteArray,
+};
+use std::ops::Deref;
+use zeroize::Zeroizing;
+
+const PRI_CONTEXT: &[u8] = b"PRIdCtx0";
+struct PseudonymousRelationshipIdentifier;
+
+impl PseudonymousRelationshipIdentifier {
```
JoeCap08055, line 20: Should we merge this struct with `DsnpPrid` already defined?

core/src/dsnp/encryption.rs
aramikm: Would you please add more details for this. We needed a concrete implementation of EncryptionBehavior to like SealBox to use the implemented traits functions encrypt. We can call EncryptionBehavior::encrypt but from my understanding we would need to pass in a Sealbox for it to know the implementation.

core/src/dsnp/pseudo_relationship_identifier.rs
aramikm: Sounds good.

core/src/dsnp/encryption.rs
JoeCap08055: nvm, I misread the code.

JoeCap08055 (commented):
lgtm

PR #10, opened 2023-03-28
feat(graph-sdk-types): implement base types for GraphSDK #4

### LINKED ISSUE #4: [Graph SDK] Implement base types
## What
Implementing based types will unblock a number of tickets to be able to continue work in parallel.

## Acceptance Criteria
- implement base types for graph sdk

### notes
You would be able to find details about these types in [here](https://github.com/LibertyDSNP/frequency/blob/main/designdocs/graph_sdk.md)

JoeCap08055:
Implementing based types will unblock a number of tickets to be able to continue work in parallel.

Acceptance Criteria
implement base types for graph sdk
notes
You would be able to find details about these types in [here](https://github.com/LibertyDSNP/frequency/blob/main/designdocs/graph_sdk.md)

Closes #4


core/src/dsnp/api_types.rs
```diff
@@ -0,0 +1,96 @@
+#![allow(dead_code)] // todo: remove after usage
+use crate::dsnp::{dsnp_types::DsnpId, encryption::EncryptionBehavior};
+
+/// Raw page of Graph (or Key) data
+pub type Page = Vec<u8>;
```
aramikm, line 5: nit: I think it would be better if we call it something else like `PageBlob` or something like that since we are going to have another `Page` struct for in memory page representation.

core/src/dsnp/api_types.rs
```diff
@@ -0,0 +1,96 @@
...
+
+pub struct KeyPair<E: EncryptionBehavior> {
```
aramikm, line 7: I'm not sure how having this as a generic might work from outside Rust boundary. I guess we can change it later if it caused problem.

core/src/dsnp/api_types.rs
```diff
@@ -0,0 +1,96 @@
...
+		connection: Connection,
+		connection_key: Option<PublicKey<E>>, // included only if PRId calculation is required
+	},
+	Disconnect {
+		owner_dsnp_user_id: DsnpId,
+		connection: Connection,
+	},
+}
+
+pub enum Update<C: Config> {
```
aramikm, line 76: I'm guessing using generic types like this would be really hard to make it compatible with C.
My assumption is that associated types makes sense if you might have different types for different  implementations but I don't think we will have different implementations for `PageId` and `SchemaId`

aramikm (approved):
Overall looks good. I added some concerns that we might need to address in future.

core/src/dsnp/api_types.rs
JoeCap08055: I agree; I think some of this will evolve with the subsequent issues & we'll adjust.

PR #11, opened 2023-03-29
Update tracker and update events implementation

### LINKED ISSUE #6: [GraphSDK] implement UpdateTracker
## what
Blocked by #4
To be able to track local changes in graph we would need to implement am `UpdateTracker`

## Acceptance Criteria
- Implement `UpdateEvent` enum or struct that represents unit changes to any graph.
- Implement `UpdateTracker` struct to keep track of these changes for different graphs and users
- complementing events should not count such as
   - Adding and Removing the same connection locally should keep the Tracker without any of those 2 events
   - Removing an existing and adding again locally should keep the Tracker without any of those 2 events
- Write tests

## Additional details
More details in  https://github.com/LibertyDSNP/frequency/blob/main/designdocs/graph_sdk.md

aramikm:
# Goal
The goal of this PR is  to add Update Tracking capability in graph which allows efficient page updates to be generated for the  chain related graph data.
closes #6 

# Acceptence Criterea
- [x] Implements UpdateEvent
- [x] Implements UpdateTracker struct to keep track of these changes
- [x] handles update complements
- [x] Write tests

# Notes
added some more documentation to types

### DIFF
--- a/core/src/dsnp/api_types.rs
+++ b/core/src/dsnp/api_types.rs
@@ -1,9 +1,10 @@
 #![allow(dead_code)] // todo: remove after usage
-use crate::dsnp::{dsnp_types::DsnpId, encryption::EncryptionBehavior};
+use crate::dsnp::{dsnp_types::DsnpUserId, encryption::EncryptionBehavior};
 
 /// Raw page of Graph (or Key) data
-pub type Page = Vec<u8>;
+pub type PageBlob = Vec<u8>;
 
+/// KeyPair used for encryption and PRI calculations
 pub struct KeyPair<E: EncryptionBehavior> {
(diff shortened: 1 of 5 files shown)


JoeCap08055 (approved):
lgtm. Only question, do we want to have `UpdateTracker` be its own struct, or as a trait that gets implemented on the `Frequency` struct? I'm fine either way, just wanted to raise it for consideration.

aramikm:
I think based on the data model we are going to have one instance of UpdateTracker per each imported user. Based on that description I think it would not be compatible with the trait on Frequency idea. Since I think Frequency is more implemented  like a singleton utility struct.

PR #13, opened 2023-04-03
feat(graph-page-structs): add structure implementations for Page and Graph

### LINKED ISSUE #5: [GraphSDK] Implement Graph and Page Structs
## Acceptance Criterea
- implement `Graph` and `Page` structs
- Each struct should be self contained and expose methods to interact with them
- Suggested methods for `Page`
   - **new**: initializing a page with page number and connections
   - **contains**: checks if a connection exists in the page
   - **is_empty**: checks if a page is empty
   - **add**: adds a connection to the page
   - **remove**: removes a connection from the page
- Suggested methods for `Graph`
   - **new**: initializing a graph with privacy level and connection types
   - **create_page**: create a page with page number and existing connections
   - **get_page**: exposes a page with page number
   - **get_connections**: returns all connections with the pages that they are in
   - **find**: returns page number of a connection if it exists in the graph and None if not
   - **add**: adds a connection to a certain page
   - **remove**: removes a connection from a certain page
- write tests

## additional info
Data model and details are located at https://github.com/LibertyDSNP/frequency/blob/main/designdocs/graph_sdk.md

JoeCap08055:
# Description
Add Page and Graph structs

# Goal
The goal of this PR is to add GraphSDK Page and Graph structs

Closes #5 

# Checklist
- [ ] Chain spec updated
- [ ] Custom RPC OR Runtime API added/changed? Updated js/api-augment.
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


core/src/dsnp/graph_page.rs
```diff
@@ -0,0 +1,215 @@
...
+	/// Add a connection to the specified page.
+	/// This is used internally by the Graph Update Manager or Import
+	/// If the specified page does not exist, a new page will be created
+	/// and the connection inserted into it.
+	pub fn add_connection_to_page<'a>(
+		&'a mut self,
+		page_id: &PageId,
+		connection_id: &DsnpUserId,
+	) -> Result<(), &str> {
+		if !self.pages.contains_key(page_id) {
```
aramikm, line 174: I think we would need to make sure that other pages don't have the same connection

core/src/dsnp/graph_page.rs
```diff
@@ -0,0 +1,215 @@
...
+		&self.prids
+	}
+
+	/// Getter for the connections in the page
+	pub fn connections(&self) -> &Vec<DsnpGraphEdge> {
+		&self.connections
+	}
+
+	/// Setter for the prids in the page
+	pub fn set_prids(&mut self, prids: Vec<DsnpPrid>) {
```
aramikm, line 53: I think we would also need add_prid for a usecase that we are adding a new connection for private graph.

Do you think maybe we need different traits for different kind of graphs? Like prid only makes sense for Private and Friendship graphs. If we can make it happen easily I'll prefer it but if it's a significant change I think we can ignore it.

core/src/dsnp/graph_page.rs
aramikm, line 1: I'm not sure if graph in-memory structs are part of DSNP. They use dsnp primitives but they might be a separate module.

core/src/dsnp/graph_page.rs
```diff
@@ -0,0 +1,215 @@
...
+
+		// Return Ok if no-op/connection not found
+		Ok(None)
+	}
+}
+
+/// Macro to get an iterator to all connections across all GraphPages
+/// within a Graph.
+#[macro_export]
+macro_rules! iter_graph_connections {
```
aramikm, line 207: I like defining macros since they are very cool. Are there any advantages with macros over having a method on graph that does the same thing?

core/src/dsnp/graph_page.rs
JoeCap08055: Hmmm, almost sounds like it might be a third module--not `dsnp`, not `frequency`, but something else?

core/src/dsnp/graph_page.rs
```diff
@@ -0,0 +1,215 @@
...
+	time::{SystemTime, UNIX_EPOCH},
+};
+
+/// Graph page structure
+#[derive(Debug, Clone, PartialEq)]
+pub struct GraphPage {
+	/// List of PRIds
+	prids: Vec<DsnpPrid>,
+	/// List of connections
+	connections: Vec<DsnpGraphEdge>,
```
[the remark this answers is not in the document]
JoeCap08055: Wasn't sure if order mattered--we were toying with the idea of keeping the `prids` in the same order as the `connections`. But `set` variants only store keys, not objects. Though I guess as long as `DsnpGraphEdge` is hashable, we can store it as a key.

core/src/dsnp/graph_page.rs
JoeCap08055: I wrote this initially as a function, but ran afoul of lifetime/ownership/sharing issues when trying to return a flattened iterator from a closure. The macro got around the issue.

core/src/dsnp/graph_page.rs
JoeCap08055: Hmm, I was thinking that prids would only be set _en masse_, either on import or a complete page refresh of prids, as we were discussing. We already have a `set_prids` method, I guess we just need an `update_prids` method, that would iterate through the connections, calculate all the prids, and set into the page. I think maybe that's the subject of another PR.

core/src/dsnp/graph_page.rs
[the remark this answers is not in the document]
JoeCap08055: Also, why `BTreeSet`? If order is not important, would `HashSet` be more efficient?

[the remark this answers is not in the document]
aramikm: I think the only difference is that for HashSet you'll need to define hashing for the struct but for BTreeSet you would only need to define `Order`. For some reason I thought defining hashing would be harder but I'm probably wrong. But you are correct that Hashset is more efficient.

core/src/dsnp/graph_page.rs
aramikm: So we are not going to calculate a separate PRid for that connection. We would need to recalculate for all others to make sure that we are not going to have stale ones.

core/src/dsnp/graph_page.rs
aramikm: yeah the naming is hard :D

JoeCap08055: I'll address this in the upcoming PRs where we'll probably do a lot of refactoring.

aramikm (approved):
Looks great. Thanks for adding the config part!

PR #16, opened 2023-04-04
Prid refactor

aramikm:
# Goal
The goal of this PR is to make sure that Prids are compatible with the test vector

related to  #9 

# Discussion
- It seems like that for **prid a->b** we were encrypting `id_a` as the message instead of `id_b` and using the other one as nonce. It does seem like in the early versions of the algorithm (at least mentioned in slack) we were doing this but we switched to the other way around. After switching to the latest version of the algorithm we are compatible with the [test vector](https://github.com/LibertyDSNP/prid-ts/blob/main/index.test.ts)
- Disabling clippy for now to speed up development process and we will enable and address comments when repo is in a more stable state
- moving other tests from their implementation files into their own file. We will need to have a decision about where is the best location for tests.
- used `u64` for `lastUpdated` since zigzag encoding takes care of efficiency and we don't have any other u32 in schema

# Checklist
- [X] Tests added

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,9 +7,4 @@ Closes <!-- issue # -->
 <!-- List discussion items -->
 
 # Checklist
-- [ ] Chain spec updated
-- [ ] Custom RPC OR Runtime API added/changed? Updated js/api-augment.
-- [ ] Design doc(s) updated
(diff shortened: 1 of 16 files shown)


JoeCap08055 (approved):
lgtm

PR #19, opened 2023-04-10
7 graphsdk implement higher level abstractions

### LINKED ISSUE #7: [GraphSDK] implement higher level abstractions
## what
Implement `User` and `ChainState` structs

## Acceptance criteria
- Implement `User` and `ChainState` which are higher layer abstraction to be able to access the graph
- implement higher level methods such as
  - importing keys and pages
  - applying a list of changes to their target graph
  - get all connections for a certain msa and graph (for persisted or dirty states) 
  - apply events to UpdateTracker
- write tests

## Additional Details
More details are defined at https://github.com/LibertyDSNP/frequency/blob/main/designdocs/graph_sdk.md

JoeCap08055:
# Goal
Implement the higher-level client-facing API for GraphSDK

Closes #7 

# Checklist
- [x] Tests added


core/src/dsnp/api_types.rs
```diff
@@ -50,17 +67,35 @@ pub trait Config {
 }
 
 /// Encapsulates all the keys and page data that needs to be retrieved from chain
-pub struct Import<E: EncryptionBehavior> {
+pub struct ImportBundle<E: EncryptionBehavior> {
```
aramikm, line 70: I think this might be useful for importing keys and pages to verify a PRI friendship but I don't think we can populate the pages and keys at the same time for private pages since we will need to decrypt connections to be able to know which public keys we will need to fetch for PRI generation

core/src/dsnp/api_types.rs
```diff
@@ -50,17 +67,35 @@ pub trait Config {
...
+	pub dsnp_user_id: DsnpUserId,
+
+	/// Connection type of graph being exported
+	pub connection_type: ConnectionType,
+
+	/// Pages to be updated in the graph
+	pub updated_pages: Vec<PageBlob>,
+
+	/// Pages to be removed from the graph
+	pub removed_pages: Vec<PageId>,
```
aramikm, line 96: I think for removing a page you would need the previous hash. And we have another update type which is **AddKey**.  This works fine but we can use enums also to specify different types. Enum would be more Rusty but for C style apis we might need to change them to something like this. So overall I don't have any preference since each option have it's own pros and cons.

core/src/dsnp/graph_page.rs
```diff
@@ -1,44 +1,158 @@
...
+	type Error = Error;
+	fn try_from(
+		(PageBlob { content_hash, payload, page_id, .. }, keys): (&PageBlob, &Vec<KeyPair<E>>),
+	) -> Result<Self> {
+		let DsnpUserPrivateGraphChunk { key_id, encrypted_compressed_private_graph, prids } =
+			DsnpUserPrivateGraphChunk::try_from(payload.as_slice())?;
+		let mut decrypted_chunk: Option<Vec<u8>> = None;
+
+		// First try the key that was indicated in the page
+		if let Some(indicated_key) = keys.iter().find(|k| k.key_id == key_id) {
```
aramikm, line 59: I think if we can not decrypt using the key mentioned we should fail the process? Right? Decryption might work with a wrong key but the decrypted data would be completely wrong and cause further problem down the line.

aramikm: The only historical checks should be done for PRI calculations

core/src/dsnp/graph_page.rs
```diff
@@ -1,44 +1,158 @@
...
+				(
+					ConnectionType::Follow(PrivacyType::Public),
+					Graph::new(ConnectionType::Follow(PrivacyType::Public)),
+				),
+				(
+					ConnectionType::Follow(PrivacyType::Private),
+					Graph::new(ConnectionType::Follow(PrivacyType::Private)),
+				),
+				(
+					ConnectionType::Friendship(PrivacyType::Public),
```
aramikm, line 134: I think we can remove this for now since we would also need to implement friendship verification

core/src/dsnp/graph_page.rs
```diff
@@ -50,12 +164,54 @@ impl UserGraph {
...
+	/// Clear the specified graph type for this user
+	pub fn clear_graph(&mut self, connection_type: &ConnectionType) {
+		if let Some(g) = self.graphs.get_mut(connection_type) {
+			g.clear();
+		}
+	}
+
+	/// Clear all graphs associated with this user
+	pub fn clear_all(&mut self) {
+		self.graphs.iter_mut().for_each(|(_, g)| g.clear());
```
aramikm, line 182: do you need to clear update_tracker also?

core/src/dsnp/graph_page.rs
```diff
@@ -100,32 +277,322 @@ impl GraphPage {
...
 		if !self.contains(connection_id) {
-			return Err("Connection not found in page")
+			return Err(Error::msg("Connection not found in page"))
 		}
 
 		self.connections.retain(|c| c.user_id != *connection_id);
 		Ok(())
 	}
+
+	// Remove all connections in the list from the page. Error if none of the connections are present in the page.
```
aramikm, line 289: the doc doesn't match to the implementation

core/src/dsnp/graph_page.rs
```diff
@@ -100,32 +277,322 @@ impl GraphPage {
...
+		}
+
+		self.set_pages(page_map);
+
+		Ok(())
+	}
+
+	/// Import bundle of pages as a Private Graph, but without decrypting the encrypted portion
+	/// Useful for importing a foreign user's graph for inspection of PRIds
+	pub fn import_opaque<E: EncryptionBehavior>(
```
aramikm, line 428: I'm thinking about a PRIManager or something like that to facilitate the PRI calculation

core/src/dsnp/graph_page.rs
```diff
@@ -100,32 +277,322 @@ impl GraphPage {
...
+				},
+			};
+		}
+
+		self.set_pages(page_map);
+
+		Ok(())
+	}
+
+	pub fn calculate_updates<E: EncryptionBehavior>(
```
aramikm, line 462: We could have implemented this in another PR but since you already done it then no complains.

core/src/dsnp/api_types.rs
JoeCap08055: The keys in the import bundle are the owner's keys needed for decryption, not other users' public keys. I'll update the doc comment to be more clear on that.

core/src/dsnp/api_types.rs
JoeCap08055: Updating...

core/src/graph/graph.rs
```diff
@@ -0,0 +1,593 @@
...
+			.collect();
+
+		// Now try to add new connections into pages already being updated
+		// Note: these pages have already been cloned, so we don't clone them again
+		let mut add_iter = ids_to_add.iter();
+		while let Some(_) = add_iter.clone().peekable().peek() {
+			if let Some((_page_id, page)) =
+				updated_pages.iter_mut().find(|(_, page)| !page.is_full(false))
+			{
+				let id_to_add = add_iter.next().unwrap();
```
aramikm, line 182: is this always safe to do here because of already checked via **`peek`**? a comment might be helpful

core/src/graph/graph.rs
```diff
@@ -0,0 +1,593 @@
...
+			if current_page.is_some() {
+				let mut page = current_page.clone().unwrap();
+				page.add_connection(add_iter.next().unwrap())?;
+				if page.is_full(false) {
+					updated_pages.insert(page.page_id(), *page);
+					current_page = None;
+				}
+			} else {
+				return Err(Error::msg("Graph is full")) // todo: re-calculate updates with agressive fullness determination
+			};
```
aramikm, line 232: This seems to be duplication of a few lines above! Copy/paste issue?

core/src/graph/page.rs
```diff
@@ -0,0 +1,411 @@
...
+		if let Some(indicated_key) = keys.iter().find(|k| k.key_id == key_id) {
+			if let Ok(data) =
+				E::decrypt(&encrypted_compressed_private_graph, &indicated_key.private_key)
+			{
+				decrypted_chunk = Some(data);
+			}
+		}
+
+		// If we couldn't decrypt with (or find) the indicated key, try all keys
+		decrypted_chunk = match decrypted_chunk {
```
aramikm, line 67: I'm still skeptical if we should try with other keys since it might be successful even though the data might be garbage. I think it depends on poly1305. Does it always fail if wrong key used to decrypt to wrong data

core/src/graph/user.rs
```diff
@@ -0,0 +1,223 @@
...
+use std::collections::HashMap;
+
+use crate::dsnp::encryption::EncryptionBehavior;
+
+use super::graph::Graph;
+
+pub type GraphMap = HashMap<ConnectionType, Graph>;
+
+/// Structure to hold all of a User's Graphs, mapped by ConnectionType
+#[derive(Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
aramikm: I think eq and partial eq should also only be checking user_id. But default derive seems to check all fields

aramikm (approved):
Awesome effort! Overall looks good. I added some comments and nits that we might be able to address before or after merge.

core/Cargo.toml
```diff
@@ -22,6 +22,7 @@ zeroize = "1.5.7"
 [dev-dependencies]
 pretty_assertions = "1.3.0"
 rand = "0.8.5"
+base64 = "*"
```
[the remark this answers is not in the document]
JoeCap08055: Ah, I threw that in there for some of the test payload stuff, didn't bother to grab a version. btw, how do you generate binary strings for byte payloads? I'd prefer not to base64 things if it's not necessary.

core/src/graph/graph.rs
```diff
@@ -0,0 +1,593 @@
...
+				updated_pages.iter_mut().find(|(_, page)| !page.is_full(false))
+			{
+				let id_to_add = add_iter.next().unwrap();
+				page.add_connection(id_to_add)?;
+			}
+		}
+
+		// Now go through the remaining connections to be added and find space in
+		// other existing pages to add them
+		let mut current_page: Option<Box<GraphPage>> = None;
```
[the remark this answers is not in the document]
JoeCap08055: Rust was giving me problems :-)

PR #20, opened 2023-04-11
added key managers

### LINKED ISSUE #17: [GraphSDK] Key input/output features
## What
Be able to import public keys as Itemized blobs and deserialize and store in memory
Create a new public keys as an Itemized blob and export it

## Acceptence criterea
- Create Interfaces and structs to import public keys as itemized blobs
- Store mentioned keys in memory
- Create interface to add a new key as an itemized blob
- Change `Update` struct mentioned in design doc to support keys
- Calculate updates should return these new added keys as an `Update`

aramikm:
# Goal
The goal of this PR is implement key handling mechanism inside the sdk

# Notes
Closes #17 
- Created `PublicKeyManager` which has a global instance which stores and retrieves dsnp public keys for graph owners and connections
- Create `UserKeyManager` which has a separate instance per each user and stores and utilizes the private keys (keypairs) for each user.
- In this PR we are not yet utilizing implemented key managers. This will be done in a future PR.


core/src/graph/key_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+pub struct PublicKeyManager {
+	/// keys are stored sorted by index
+	dsnp_user_to_keys: HashMap<DsnpUserId, (Vec<DsnpPublicKey>, PageHash)>,
+	new_keys: HashMap<DsnpUserId, DsnpPublicKey>,
+}
+
+pub struct UserKeyManager {
+	public_key_manager: RwLock<PublicKeyManager>,
```
aramikm, line 39: This is giving me a hard time. I tried a Rc and RefCell and .. but no success. I guess I have to reread all those.

The idea is to have one instance of PublicKeyManager and share that reference with UserKeyManagers. UserKeyManager should be only reading from it.

core/src/graph/key_manager.rs
```diff
@@ -0,0 +1,283 @@
...
+
+pub struct UserKeyManager {
+	public_key_manager: Rc<RefCell<PublicKeyManager>>,
+	dsnp_user_id: DsnpUserId,
+	keys: Vec<StackKeyPair>,
+}
+
+impl PublicKeyProvider for PublicKeyManager {
+	/// importing dsnp keys as they are retrieved from blockchain
+	/// sorting indices since id's might not be unique but indices definitely should be
```
JoeCap08055, line 46: Hmm... I'm wondering if it might be worth building an item key into Itemized storage, instead of just relying on positional index...
* Wouldn't really be any more computationally expensive, since we're already reading the items into a BTreeMap indexed by item (would just change index to key)
* Wouldn't involve any different storage reads/writes
* Would be able to eliminate duplicate-keyed items on-chain
* Could support "upsert" for items

Drawbacks:
* Would not be able to "update" an item in AppendOnly by re-publishing

core/src/graph/key_manager.rs
```diff
@@ -0,0 +1,283 @@
...
+		}
+		if let Some(k) = self.new_keys.get(&dsnp_user_id) {
+			all_keys.push(k)
+		}
+		all_keys
+	}
+
+	fn get_key_by_id(&self, dsnp_user_id: DsnpUserId, key_id: u64) -> Option<&DsnpPublicKey> {
+		// What should we do if there are more than one key with the same key_id uploaded?
+		// currently choosing the first one
```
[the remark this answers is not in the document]
aramikm: I'll add a utility function to get the valid key from an iterater

core/src/dsnp/api_types.rs
```diff
@@ -126,8 +149,8 @@ pub enum Update {
 		payload: Vec<u8>,
```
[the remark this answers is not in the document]
aramikm: We decided to remove the schema id from execution path is sdk and in future we will be able to expose it as metadata for the upper level to use it

core/src/dsnp/api_types.rs
```diff
@@ -166,17 +189,32 @@ pub enum Update {
...
+		connection_type: ConnectionType,
+
 		/// page id associated with changed page
 		page_id: PageId,
 
 		/// previous hash value is used to avoid updating a stale state
 		prev_hash: Vec<u8>,
 	},
+
+	/// A `Delete` type is used to remove a page from the chain
```
JoeCap08055, line 207: nit: update comment

aramikm: Updated

PR #32, opened 2023-04-13
Using key managers

### LINKED ISSUE #30: [GraphSDK] Integrate Key managers with current implemementation
# What
Key managers are implemented in # and now we need to utilize them.

# Acceptence Creterea
- Use Public key manager to store all public keys
- Use User key manager to store user key pairs coming from wallet

aramikm:
# Goal
The goal of this PR is to use implemented key managers

Closes #30 

# Discussion
- Adding Builders for entities to allow easier test setup

# Checklist
- [x] Tests added


core/src/graph/graph.rs
```diff
@@ -57,9 +57,10 @@ impl Graph {
 	}
 
 	/// Import bundle of pages as a Public Graph
-	pub fn import_public<E: EncryptionBehavior>(
```
aramikm, line 60: It seems like we might need something like a Multihash that specifies the key type. Or we might need the encryption type from spec config.

core/src/graph/page.rs
```diff
@@ -57,7 +57,7 @@ impl<E: EncryptionBehavior> TryFrom<(&PageData, &Vec<KeyPair<E>>)> for GraphPage
 		// First try the key that was indicated in the page
 		if let Some(indicated_key) = keys.iter().find(|k| k.key_id == key_id) {
 			if let Ok(data) =
-				E::decrypt(&encrypted_compressed_private_graph, &indicated_key.private_key)
+				SealBox::decrypt(&encrypted_compressed_private_graph, &indicated_key.key_pair)
```
aramikm, line 60: replaced with concrete implementation but I think we should decide about which way we want to go based on previous comment.

core/src/api/api.rs
```diff
@@ -118,24 +128,37 @@ impl<E: EncryptionBehavior, const M: usize> GraphAPI<E> for GraphState<E, M> {
...
-					graph.import_opaque(payload)
-				} else {
-					graph.import_private(payload)
-				},
+		match (connection_type.privacy_type(), include_secret_keys) {
+			(PrivacyType::Public, _) => graph.import_public(connection_type, pages),
+			(PrivacyType::Private, true) =>
+				graph.import_private(connection_type, pages, resolved_keys),
+			// TODO: import into PRId manager
+			(PrivacyType::Private, false) => graph.import_opaque(connection_type, pages),
```
JoeCap08055, line 161: 💯

core/src/api/api.rs
```diff
@@ -7,19 +7,23 @@ use crate::{
...
 use anyhow::{Error, Result};
-use std::{cmp::min, collections::HashMap, marker::PhantomData};
+use std::{cell::RefCell, cmp::min, collections::HashMap, marker::PhantomData, rc::Rc};
 
 const MAX_GRAPH_USERS_DEFAULT: usize = 1000;
 
-#[allow(non_snake_case)]
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug)]
 pub struct GraphState<E: EncryptionBehavior, const MAX_USERS: usize = MAX_GRAPH_USERS_DEFAULT> {
```
JoeCap08055, line 24: Based on your changes, I think we can completely get rid of the `<E: EncryptionBehavior>` generic parameter, pending outcome of discussion of the points you raised below.

aramikm: I think so, will do that in another PR that standardizes apis

JoeCap08055 (approved):

PR #33, opened 2023-04-14
feat(refactor configs): refactor config into separate crate, and fix infinite loop bug in export

JoeCap08055:
Refactor configuration data (chain constants, schema ids & schemas) into their own independent crate

# Checklist
- [x] Tests added


config/resources/configs/frequency.json
```diff
@@ -0,0 +1,35 @@
+{
+  "maxGraphPageSizeBytes": 2048,
```
aramikm, line 2: I think this should be 1024

config/resources/configs/frequency.json
```diff
@@ -0,0 +1,35 @@
...
+  "maxPageId": 16,
+  "maxKeyPageSizeBytes": 1024,
```
aramikm, line 4: This is higher around 64k

config/resources/configs/frequency-rococo.json
```diff
@@ -0,0 +1,35 @@
+{
+  "maxGraphPageSizeBytes": 2048,
```
aramikm, line 2: 1024

config/resources/configs/frequency-rococo.json
```diff
@@ -0,0 +1,35 @@
...
+  "maxPageId": 16,
+  "maxKeyPageSizeBytes": 1024,
```
aramikm, line 4: 64k if I'm not wrong

config/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+			.unwrap();
+	pub static ref PUBLIC_GRAPH_SCHEMA: Schema =
+		Schema::parse_str(include_str!("../resources/schemas/public_graph.json")).unwrap();
+	pub static ref PRIVATE_GRAPH_CHUNK_SCHEMA: Schema =
+		Schema::parse_str(include_str!("../resources/schemas/user_private_graph_chunk.json"))
+			.unwrap();
+}
+
+/// Privacy Type of the graph
+#[derive(Clone, Copy, PartialEq, Ord, Eq, PartialOrd, Debug, Hash, Serialize, Deserialize)]
```
aramikm, line 26: Should these types be defined out side config? any intentions why here?

core/src/graph/graph.rs
```diff
@@ -603,4 +609,17 @@ mod test {
 		graph_connections.sort();
 		assert_eq!(test_connections, graph_connections);
 	}
+
+	#[test]
+	#[timeout(5000)] // found an infinite loop bug, so let's make sure this terminates successfully
```
aramikm, line 614: interesting, should check our loop conditions

config/resources/configs/frequency-rococo.json
JoeCap08055: right; I put in placeholders & then forgot to check in the frequency repo 🤦🏻

config/src/lib.rs
JoeCap08055: In order to be able to have config manage the mapping ConnectionType <=> SchemaId, it needs to know about those enums. If we define them in the core crate, then we get a circular dependency. The other option would be to factor out into a `dsnp-graph-common` or `dsnp-graph-types` crate...

aramikm: I see, let's keep it here then.

config/resources/configs/frequency-rococo.json
JoeCap08055: Updated to match frequency repo

config/resources/configs/frequency.json
JoeCap08055: Updated to match frequency repo

config/resources/configs/frequency.json
JoeCap08055: Updated to match frequency repo

aramikm (approved):
Looks good!

PR #36, opened 2023-04-19
use schema ids in apis

aramikm:
# Goal
The goal of this PR is address schema_id oriented graph and api structures

related to #35 

# Discussion
- Using schema ids instead of connection types
- add support for different environments

# Checklist
- [x] Fix Tests

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -1,3 +1,5 @@
+mod builder;
+
 use apache_avro::Schema;
 use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
@@ -61,10 +63,11 @@ impl ConnectionType {
 	}
 }
 
-#[derive(Clone, PartialEq, Ord, Eq, PartialOrd, Debug)]
+#[derive(Clone, PartialEq, Eq, Debug)]
 pub enum Environment {
 	Mainnet,
 	Rococo,
+	Dev(Config),
 }
 
 impl Environment {
@@ -72,6 +75,7 @@ impl Environment {
 		match self {
 			Environment::Mainnet => &MAINNET_CONFIG,
 			Environment::Rococo => &ROCOCO_CONFIG,
+			Environment::Dev(cfg) => &cfg,
 		}
 	}
 }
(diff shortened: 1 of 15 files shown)


config/src/lib.rs
```diff
@@ -57,6 +61,48 @@ impl ConnectionType {
 	}
 }
 
+#[derive(Clone, PartialEq, Ord, Eq, PartialOrd, Debug)]
+pub enum Environment {
+	Mainnet,
+	Rococo,
+}
```
JoeCap08055, line 68: I think we also should have: `Dev(Config)` for ease of testing.

JoeCap08055 (commented):
Looks mostly good. However, I don't think we need to be passing `Environment` down into all of the lower-level user/graph/page objects. Main use case is there will only ever be a single Environment for an application... can we make it a static at the crate level or something?

core/src/api/api.rs
```diff
@@ -269,38 +270,40 @@ mod test {
...
-		let new_state = TestGraphState::<TEST_CAPACITY>::with_capacity(TEST_CAPACITY * 2);
+		let new_state =
+			TestGraphState::<TEST_CAPACITY>::with_capacity(Environment::Mainnet, TEST_CAPACITY * 2);
 		assert!(new_state.user_map.capacity() >= TEST_CAPACITY);
 	}
 
 	#[test]
 	fn graph_state_capacity() {
-		let state: TestGraphState = TestGraphState::new();
+		let state: TestGraphState = TestGraphState::new(Environment::Mainnet);
```
JoeCap08055, line 287: lol, this scared me to reference "Mainnet" in a test suite, until I remembered that Mainnet config has nothing to do with communicating on-chain; it's just for constants & schema definition. Still scary to read when your eye first scans it, though 🤷🏻

JoeCap08055 (approved):

core/src/api/api.rs
aramikm: Yeah we would need some more tests with some non mainnet configs to make sure everything works as expected but will do that in another PR.

PR #37, opened 2023-04-20
added pri manager

### LINKED ISSUE #31: [GraphSDK] Create PRI Manager
# what
To be able to calculate PRIs across all users and connection we will need a global instance of a struct to store all imported PRIs for msa ids

## Acceptence Criterea
- PRI manager should be able import PRI's without the need to decrypt the page
- Store all PRI stored on a msa id
- Keep the PRId in the same order as connections stored in the page

aramikm:
# Goal
The goal of this PR is implement PRI Manager which stores and retrieves all imported PRId to facilitate prid calculations and checks

Closes #31

### DIFF
--- a/core/src/api/api.rs
+++ b/core/src/api/api.rs
@@ -6,6 +6,7 @@ use crate::{
 	},
 	graph::{
 		key_manager::{PublicKeyManager, PublicKeyProvider, UserKeyProvider},
+		pri_manager::{PriManager, PriProvider},
 		updates::UpdateEvent,
 		user::UserGraph,
 	},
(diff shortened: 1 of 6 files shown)


core/src/api/api.rs
```diff
@@ -23,6 +24,7 @@ pub struct GraphState<E: EncryptionBehavior, const MAX_USERS: usize = MAX_GRAPH_
 	phantom: PhantomData<E>,
 	environment: Environment,
 	public_key_manager: Rc<RefCell<PublicKeyManager>>,
+	pri_manager: Rc<RefCell<PriManager>>,
```
JoeCap08055, line 27: This has me thinking; should we adopt this approach for the Environment that gets passed down into all of the graphs? Thinking that could be a bit heavyweight due to holding multiple complete Avro schemas in each copy. Maybe something to file a task for later.

JoeCap08055 (approved):

core/src/api/api.rs
aramikm: Since we only used the reference to config in the environment. The overhead is minuscule.

PR #39, opened 2023-04-25
modified apis to support prids and batch

### LINKED ISSUE #23: [graphSDK] Standardize API to support PRI
# Problem
Standardize and modify APIs to support PRI calculation workflows

## Acceptence Criterea
- APIs should be able to allow PRI **verification** and **generation** for each page in Private Friendship graph
- Expose which connections require public keys

aramikm:
# Goal
The goal of this PR is modify apis to support PRID usecase and allow batch processing

Closes #23 

# Discussion
- Tests are not fixed yet
- Batch apis would need transactional support

# Checklist
- [x] Tests added

### DIFF
--- a/core/src/api/api.rs
+++ b/core/src/api/api.rs
@@ -234,9 +234,9 @@ impl GraphAPI for GraphState {
 				};
 
 				owner_graph.update_tracker_mut().register_update(&update_event)?;
+			} else {
+				return Err(Error::msg("user graph not found in state"))
 			}
-
-			return Err(Error::msg("user graph not found in state"))
 		}
 		Ok(())
 	}
@@ -283,10 +283,9 @@ impl GraphAPI for GraphState {
 		let all_connections: HashSet<DsnpUserId> = self
 			.user_map
 			.values()
-			.map(|user_graph| {
+			.flat_map(|user_graph| {
 				user_graph.get_all_connections_of(ConnectionType::Friendship(PrivacyType::Private))
 			})
-			.flatten()
 			.collect();
 		self.public_key_manager
 			.deref()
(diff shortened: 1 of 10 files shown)


core/src/api/api.rs
```diff
@@ -135,107 +134,111 @@ impl GraphAPI for GraphState {
...
+				false => vec![],
+			};
 
-		match (connection_type.privacy_type(), include_secret_keys) {
-			(PrivacyType::Public, _) => graph.import_public(connection_type, pages),
-			(PrivacyType::Private, true) => {
-				graph.import_private(&dsnp_config, connection_type, &pages, resolved_keys)?;
-				self.pri_manager
+			let graph = user_graph.graph_mut(&schema_id);
+			graph.clear();
```
JoeCap08055, line 170: Regarding the TODO above to make this transactional, I suppose one step toward that would be to create a new graph and only replace the existing one when the import completes successfully.

core/src/graph/key_manager.rs
```diff
@@ -176,12 +179,27 @@ impl PublicKeyProvider for PublicKeyManager {
 
 		Ok(())
 	}
+
+	fn filter_users_without_keys(&self, dsnp_user_ids: Vec<DsnpUserId>) -> Vec<DsnpUserId> {
```
JoeCap08055, line 183: nit: maybe 'get' or 'find' would be more appropriate naming than 'filter'

core/src/api/api.rs
aramikm: We can do that or I was thinking about having a dryrun flag described in https://github.com/LibertyDSNP/graph-sdk/issues/40. Basically we run all the input using **dryrun = true** and if everything looked good we would run it again.

core/src/graph/key_manager.rs
aramikm: will change

core/src/api/api.rs
aramikm: Open to both ideas. We can do a quick evaluation of how much is the effort with each one and progress from there. I'll update that ticket to have your approach as an alternative to explore.

JoeCap08055 (approved):

Issue #42, opened 2023-04-27
[GraphSDK] comprehensive integration tests


p5150j:
We want to add some comprehensive integration tests. Blackbox testing.

PR #50, opened 2023-05-02
implemented api to expose one sided friendships

### LINKED ISSUE #48: [GraphSDK] Add a new api endpoint to expose one sided friendships
## Problem
We would like to allow providers to know more about one sided friendships in graph.

# Acceptence Criterea
- add an api that would expose connections on private friendships types that are one sided and their connection establishment timestamp.
- add tests

# Notes
Can look into `applyPrids` on **graph.r**s to show how we calculate stale one sided connections.

aramikm:
# Goal
The goal of this PR is to implement an api so that we can return one sided friendships to sdk consumers.

Closes #48 

# Discussion
- Implemented a new `MockUserKeyManager` to be able to setup tests more easily. 
- Added support to be able to use mock in for `UserKeyManager`

# Checklist
- [x] Tests added

### DIFF
--- a/core/src/api/api.rs
+++ b/core/src/api/api.rs
@@ -60,6 +60,12 @@ pub trait GraphAPI {
 
 	/// return a list dsnp user ids that require keys
 	fn get_connections_without_keys(&self) -> Result<Vec<DsnpUserId>>;
+
+	/// Get a list of all private friendship connections that are only valid from users side
+	fn get_one_sided_private_friendship_connections(
+		&self,
+		user_id: &DsnpUserId,
+	) -> Result<Vec<DsnpGraphEdge>>;
 }
 
 impl GraphState {
@@ -235,6 +241,7 @@ impl GraphAPI for GraphState {
 		Ok(user_graph.get_all_connections_of(*schema_id, include_pending))
 	}
 
+	/// return a list dsnp user ids that require keys
 	fn get_connections_without_keys(&self) -> Result<Vec<DsnpUserId>> {
 		let private_friendship_schema_id = self
 			.environment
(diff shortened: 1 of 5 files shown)


core/src/graph/graph.rs
```diff
@@ -28,7 +27,7 @@ pub struct Graph {
 	user_id: DsnpUserId,
 	schema_id: SchemaId,
 	pages: PageMap,
-	user_key_manager: Rc<RefCell<UserKeyManager>>,
+	user_key_manager: Rc<RefCell<dyn UserKeyManagerBase>>,
```
aramikm, line 30: Allows injecting Mock. Didn't use generic `Graph<E>` since it the code change would be much more than desired.

core/src/api/api.rs
```diff
@@ -255,6 +262,24 @@ impl GraphAPI for GraphState {
 			.borrow()
 			.find_users_without_keys(all_connections.into_iter().collect()))
 	}
+
+	/// Get a list of all private friendship connections that are only valid from users side
+	fn get_one_sided_private_friendship_connections(
```
aramikm, line 267: Majority of these api tests should be moved into integration tests and didn't implement a new test in api for this since it would be an integration test setup and we will do it in #42 for all apis

JoeCap08055: I was thinking the same thing; most of the top-level `api` tests are integration-level. 👍🏻

core/src/graph/graph.rs
JoeCap08055: 💯

core/src/graph/graph.rs
```diff
@@ -411,13 +413,34 @@ impl Graph {
 		Ok(None)
 	}
 
+	/// returns one sided friendship connections
+	pub fn get_one_sided_friendships(&self) -> Result<Vec<DsnpGraphEdge>> {
+		if self.get_connection_type() != ConnectionType::Friendship(PrivacyType::Private) {
+			return Err(Error::msg(
+				"Calling get_all_one_sided_friendships in non private friendship graph!",
+			))
+		}
```
JoeCap08055, line 422: Technically, a public friendship could also be one-sided, it just doesn't require PRId.

JoeCap08055 (approved):
Looks good. Main question: do we want to support fetching one-sided Public friendships as well? Would mean we probably couldn't use the key manager to verify, but instead would have to go right to the other users' Public graphs...

Approved; if we think we should support that, let's make another issue.

core/src/graph/graph.rs
aramikm: I don't think we have any public friendship use-case currently but will create a ticket to track it in case it comes up in future.

PR #51, opened 2023-05-03
added support for time restricted prid connection removal

### LINKED ISSUE #47: [GraphSDK] Time restricted PRId friendship removal
## Problem
Currently we are checking the PRID existence and remove any connections immediately if we can not detect the friendship from other side. This might cause some race condition problems and we would like to only remove the connection if a certain amount of time (like 3 months is passed) from connection establishment time.

## Acceptance Criterea
- Add a config in our json config files that will have the time period needed to remove a stale friendship. Might be able to name it `allowedStaleFriendshipPeriod` or something else.
- Use that mentioned config inside `apply_prids` function inside **graph.rs** and only remove connections that are older than mentioned period.
- Add/Modify tests to cover changed logic.

aramikm:
# Goal
The goal of this PR is only remove friendship connections only if the friendship establishment is passes a certain time (90 days) and the expected PRID does not exist on the other side. 

Closes #47 

# Checklist
- [x] Tests added/Modified

### DIFF
--- a/core/src/dsnp/encryption.rs
+++ b/core/src/dsnp/encryption.rs
@@ -77,7 +77,7 @@ mod test {
 
 		let key_pair = KeyPairType::Version1_0(StackKeyPair::from_seed(&[0, 1, 2, 3, 4]));
 		let mut encrypted = SealBox.encrypt(&plain_data, &key_pair.borrow().into()).unwrap();
-		encrypted[1] = encrypted[1].saturating_add(1); // corrupting data
+		encrypted[1] = !encrypted[1]; // corrupting data
 		let decrypted = SealBox.decrypt(&encrypted, &key_pair.into());
 
 		assert!(decrypted.is_err());
--- a/core/src/util/time.rs
+++ b/core/src/util/time.rs
@@ -1,6 +1,40 @@
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-/// Calculate a timestamp in seconds, rounded to the nearest 1000
+const DAYS: u64 = 60 * 60 * 24;
+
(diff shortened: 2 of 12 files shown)


core/src/dsnp/encryption.rs
```diff
@@ -77,7 +77,7 @@ mod test {
 
 		let key_pair = KeyPairType::Version1_0(StackKeyPair::from_seed(&[0, 1, 2, 3, 4]));
 		let mut encrypted = SealBox.encrypt(&plain_data, &key_pair.borrow().into()).unwrap();
-		encrypted[1] = encrypted[1].saturating_add(1); // corrupting data
+		encrypted[1] = !encrypted[1]; // corrupting data
```
aramikm, line 80: Fixing flaky tests

saraswatpuneet (approved):
lgtm!

core/src/dsnp/encryption.rs
JoeCap08055: Ugh. I don't like the fact that Rust overloads `!` for both bitwise and boolean NOT 😞 
Oh well.

core/src/util/time.rs
```diff
@@ -1,6 +1,40 @@
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-/// Calculate a timestamp in seconds, rounded to the nearest 1000
+const DAYS: u64 = 60 * 60 * 24;
```
JoeCap08055, line 3: Maybe `SECS_PER_DAY` would be a better name?
Or we could use this: https://docs.rs/arrow2/latest/arrow2/temporal_conversions/constant.SECONDS_IN_DAY.html
(though it seems ridiculous to add a dependency just for time constants)

aramikm: Will rename it in my future PR.

PR #54, opened 2023-05-05
added full transactional operation support

### LINKED ISSUE #40: [GraphSDK] Add transactional support to batch apis
## What
To be able to keep the state clean in the graph sdk we will need to add transactional support to batch api endpoints.

# Acceptance Criteria
- add a **dryrun** flag in environment or some other place that gets injected into all other components and if it's enabled the state should should not change
- first run all the batch inputs in **dryrun** mode and if nothing fails then rerun everything to apply the state.

## Alternative Approach
create a new graph and only replace the existing one when the import completes successfully.

Need a quick flexibility and effort assessment for each approach and choose the lower effort.

aramikm:
# Goal
The goal of this PR is support full in memory transactional operation support

Closes #40 

# Discussion
- Created two new Types `TransactionalHashMap` and `TransactionalVec` which tracks changes optimistically. And in case of rollback to previous state.
- in case of success or failure the highest component `Api` triggers a `commit` or `rollback` which gets propagated down into the lower components 

# Checklist
- [x] Tests added

### DIFF
--- a/core/src/api/api.rs
+++ b/core/src/api/api.rs
@@ -124,7 +124,7 @@ impl GraphAPI for GraphState {
 
 	/// Return number of users in the current graph state
 	fn len(&self) -> usize {
-		self.user_map.inner().len()
+		self.user_map.len()
 	}
 
 	/// Remove the user graph from an instance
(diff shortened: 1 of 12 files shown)


JoeCap08055 (approved):
Very cool! LGTM. Though after all that effort it has me wondering if we shouldn't have simply used something like sqlite in-memory db to keep the graph state 😅

aramikm:
I'm not sure if having a dependency on an in-memory db that is graph compatible and will work across all platforms would be easier in the long run.

PR #56, opened 2023-05-05
feat(#25): Add calculations for lazy page fullness with worst-case compression

### LINKED ISSUE #25: [GraphSDK] Spike to calculate number of connection in a page
# Problem
For non aggressive algorithm we would like to have the minimum number of connections that can fit in graph pages.

## Acceptence Criterea
- Calculate minimum connection for Private Friendship graph page
- Calculate minimum connection for Private Follow graph page
- Calculate minimum connection for public graph page

JoeCap08055:
# Goal
The goal of this PR is to add the ability to calculate the max # of connections/page for each connection type using worst-case compression scenarios.

* Add. ability to run a test that will auto-generate a capacity file.

Closes #25 

# Checklist
- [x] Tests added


config/Cargo.toml
```diff
@@ -16,3 +21,6 @@ apache-avro = { version = "0.14.0", features = ["snappy"] }
 
 [dev-dependencies]
 pretty_assertions = "1.3.0"
+
+[features]
+calculate-page-capacity = []
```
aramikm, line 26: How do you think about maybe putting all these calculation logic in `bench` folder? Since this seems like a benchmark if we do that then I think we can get away with having an extra feature for this. In general I'm a bit conservative towards using features unless absolutely necessary (like for wasm) and prefer alternative approaches.

config/src/lib.rs
```diff
@@ -166,7 +166,7 @@ impl Config {
 	}
 }
 
-#[cfg(test)]
+#[cfg(all(test, not(feature = "calculate-page-capacity")))]
```
aramikm, line 169: As mentioned, this might get messy real quick.

config/Cargo.toml
JoeCap08055: We can certainly separate the code out into a separate folder; however, the bigger issue is I wanted to be able to run the graph calculation without having to run all the other tests. But the important part is not running the graph calculation every time we run tests, which is achieved with only the one `cfg` statement around the graph calculation module. I say we leave the rest of these for now; if it gets messy to maintain we can remove them later.

I initially tried to set up the graph capacity calculation as an actual benchmark, but apparently the `#[bench]` attribute is currently unstable & only available in nightly. So I had to resort to setting it up as a test.

The other alternative would be to adopt the Substrate approach to benchmarking, which would be to build a binary and give it a CLI mode to run the calculation. But that seemed like overkill.

aramikm: Another alternative approach might be yo just `ignore` that benchmarking test just like parity does https://github.com/paritytech/substrate/blob/master/frame/message-queue/src/integration_test.rs#L125
I think that would be cleaner compared to adding feature flag on all test location. Any thoughts?

core/src/benches/page_size.rs
```diff
@@ -0,0 +1,80 @@
...
+		let mut result_vec: Vec<usize> = Vec::new();
+		for _ in 0..1000 {
+			result_vec.push(benchmark_page_capacity(c).0);
+		}
+		result_vec.sort();
+		capacity_map.insert(c, *result_vec.first().unwrap());
+	}
+
+	let code = format!(
+			"use dsnp_graph_config::{{ConnectionType, ConnectionType::Follow, ConnectionType::Friendship, PrivacyType::Public, PrivacyType::Private}};
```
aramikm, line 57: Should this have some auto-generated comment som that people know not to modify it?

aramikm (approved):
Overall looks good! Added a couple of comments/suggestions that is not a blocker for this PR.

config/Cargo.toml
JoeCap08055: The issue isn't how to ignore the benchmarking test--that's easy and only requires an attribute in one spot. The issue is I want to be able to ignore all the other tests when we want to run the benchmark. To do that we need to flag all the other test modules. Unless you thought of an easier way?

One other option would be to forget about all that, and just control it at the Makefile level like this:

```
calculate-page-capacities:
     cargo test --feature=calculate_page_capacities calculate_page_capacities
```

The last argument is a string matcher for the test(s) to run.

If you prefer that, let me know; it's easy enough to change.

aramikm: I honestly prefer the last mentioned option compared to adding feature flags to all tests. The reason is that this is just needs to be ran every once in a while if not only once.

Issue #28, opened 2023-04-12
[GraphSDK] better errors


p5150j:
Currently errors are lacking deep context, we would like to add stack trace or more verbose "useable" errors and.
- More FFI friendly errors

PR #57, opened 2023-05-05
C compatible FFI for graph sdk

### LINKED ISSUE #15: [GraphSDK] implement C compatible interface of sdk
## What
To be able to interact with other languages we will need **c** style api's.

saraswatpuneet:
# Goal
The goal of this PR is  to implement c compatible ffi for graph sdk

Closes #15 

# Checklist
- [x] Tests added


bridge/ffi/cbindgen.toml
```diff
@@ -0,0 +1,23 @@
+# cbindgen for generating C bindings
+language = "C"
+header = "/*\n Copyright (C) 2023 Frequency Team /\n */"
+autogen_warning = "/* This file was automatically generated by cbindgen */"
+include_guard = "GRAPH_SDK_FFI_H_"
+style = "type"
+
+[export]
+include = ["Config", "GraphState", "Environment", "DsnpVersion", "ConnectionType", "PrivacyType", "SchemaConfig","SchemaConfigTuple", "ImportBundle", "Update", "Action", "DsnpGraphEdge"]
```
aramikm, line 9: I didn't know we have to include all items manually.

bridge/ffi/src/bindings.rs
```diff
@@ -0,0 +1,203 @@
+use dsnp_graph_config::{ConnectionType, DsnpVersion, GraphKeyType, SchemaId};
+use dsnp_graph_core::dsnp::{
+	api_types::{Connection, PageHash, PageId},
+	dsnp_types::DsnpUserId,
+};
+use libc::{c_void, size_t};
+/// KeyData wrapper
+#[repr(C)]
+pub struct KeyData {
```
aramikm, line 9: If the Rust type is supported in C representation. Should we just use that `#[repr(C)]` instead of redefining all? What are the pros and cons of each choice?

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,34 @@
...
+use libc::c_void;
+
+#[no_mangle]
+pub extern "C" fn print_hello_graph() {
+	println!("Hello, Graph!");
+}
+
+// Define a C-compatible representation of GraphState
+#[repr(C)]
+pub struct GraphState {
```
aramikm, line 12: I wasn't sure we would need the GraphState defined. I know that we would need to keep the state somewhere and we might need to create a singleton one.

bridge/ffi/src/bindings.rs
saraswatpuneet: Vec is not supported, so we would need them as far as i found so far, but will clean if possible

bridge/ffi/src/bindings.rs
```diff
@@ -0,0 +1,203 @@
...
+	index: u16,
+	content: *mut u8,
+}
```
saraswatpuneet, line 12: probably will need another field for content_len

bridge/ffi/src/c_api.rs
saraswatpuneet: We would need c types for GraphState since it is comprised of non primitive types like Vec HashMap, however since user of the api wont directly interact with state but via exposed function, we would just keep a reference to state in C type, encapsulating it

saraswatpuneet: But think a singleton would solve it, so good recommendation

bridge/ffi/src/bindings.rs
saraswatpuneet: We can also use feature flag but that will make rust code look ugly

bridge/ffi/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "graph-sdk-ffi"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+publish = false
+
+[lib]
+name = "graph_sdk_ffi"
+crate-type = ["cdylib"]
```
saraswatpuneet, line 10: Need static or dynamic lib or both?

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,188 @@
...
+//Graph import users data
+#[no_mangle]
+pub unsafe extern "C" fn graph_import_users_data(
+	payloads: *const ImportBundle,
+	payloads_len: usize,
+) -> bool {
+	if let Some(graph_state) = GRAPH_STATE.as_mut() {
+		let payloads = std::slice::from_raw_parts(payloads, payloads_len);
+		let payloads = payloads_from_ffi(&payloads);
+		graph_state.import_users_data(&payloads).is_ok()
```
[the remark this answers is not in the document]
aramikm: I think this might be related to Error cleanup ticket #28  also since the errors that we expose might need to be mapped to an integer and having enum errors (with an int value) might help with that.

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,188 @@
...
+// Graph export updates
+#[no_mangle]
+pub unsafe extern "C" fn graph_export_updates() -> GraphUpdates {
+	if let Some(graph_state) = GRAPH_STATE.as_mut() {
+		match graph_state.export_updates() {
+			Ok(updates) => {
+				let mut ffi_updates = updates_to_ffi(updates);
+				let updates_len = ffi_updates.len();
+				let updates_ptr = ffi_updates.as_mut_ptr();
+				std::mem::forget(ffi_updates);
```
aramikm, line 107: Following statement [in ](https://doc.rust-lang.org/std/mem/fn.forget.html)worries me a bit. Is it possible to use `ManuallyDrop` instead?

> Any resources the value manages, such as heap memory or a file handle, will linger forever in an unreachable state.

aramikm (approved):
Looks good to me! I added some nits and comments but prefer @JoeCap08055 to also take another look since I have't used FFI and C for a while.

bridge/ffi/src/c_api.rs
saraswatpuneet: The consuming application would get the pointer to that memory address `updates_ptr` hoping downstream consumers cleanup?

saraswatpuneet: Or we want to collect at our end and upon call to free the graph we also cleanup any memory footprint we created?

bridge/ffi/Cargo.toml
saraswatpuneet: Looks like we need static , I have made the change

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,188 @@
...
+		let payloads = payloads_from_ffi(&payloads);
+		graph_state.import_users_data(&payloads).is_ok()
+	} else {
+		false
+	}
+}
+
+// Graph export updates
+#[no_mangle]
+pub unsafe extern "C" fn graph_export_updates() -> GraphUpdates {
```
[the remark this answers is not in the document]
saraswatpuneet: I think that's a good idea, we can define a high level return type C_Result that has data and error field or something like that

bridge/ffi/src/c_api.rs
saraswatpuneet: I read that, it worries me too,would check the alternative

saraswatpuneet: > This is useful when the ownership of the underlying resource was previously transferred to code outside of Rust, for example by transmitting the raw file descriptor to C code.

saraswatpuneet: yep ManuallyDrop is better option

Makefile
```diff
@@ -37,4 +37,10 @@ clean:
 all: check test clippy fmt build doc
 
 .PHONY: ci-local
-ci-local: all
\ No newline at end of file
+ci-local: all
+
+bindgen:
+	@echo "Running bindgen..."
+	@cargo install cbindgen
```
JoeCap08055, line 44: This seems like a code smell to me... installing packages should be a part of the runner setup in CI, and a documented manual step for local development, since it only needs to be done once per machine.

bridge/ffi/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "graph-sdk-ffi"
```
JoeCap08055, line 2: Can we use consistent naming across our GraphSDK crates? Other crates are `dsnp-graph-*`; how about `dsnp-graph-ffi`?

bridge/ffi/src/c_example/graph_sdk_ffi.h
```diff
@@ -0,0 +1,390 @@
+/*
```
JoeCap08055, line 1: Do we want to commit this file, since it's generated?

bridge/ffi/src/utils.rs
```diff
@@ -0,0 +1,180 @@
...
+) -> Vec<dsnp_graph_core::dsnp::api_types::ImportBundle> {
+	let mut rust_payloads = Vec::new();
+	for payload in payloads {
+		let rust_payload = import_bundle_from_ffi(payload);
+		rust_payloads.push(rust_payload);
+	}
+	rust_payloads
+}
+
+// Function to convert C-compatible `Update` to a Rust `Update`
```
JoeCap08055, line 128: Shouldn't this comment be reversed?
```
// Function to convert Rust `Update` to C-compatible `Update`
```

bridge/ffi/src/utils.rs
```diff
@@ -0,0 +1,180 @@
...
+				page_id,
+				prev_hash,
+				mut payload,
+			} => {
+				let ffi_persist_page = PersistPage {
+					owner_dsnp_user_id: owner_dsnp_user_id.clone(),
+					schema_id,
+					page_id,
+					prev_hash,
+					payload: payload.as_mut_ptr(),
```
JoeCap08055, line 145: How do we guarantee the original `payload` will not be cleaned up? Don't we need clone + `ManuallyDrop` or something? Same again below.

bridge/ffi/src/utils.rs
```diff
@@ -0,0 +1,180 @@
+use crate::bindings::*;
```
JoeCap08055, line 1: For these `*_from_ffi` and `*_to_ffi` functions, could we implement as traits?
```
pub trait ToFFI<RustType> {
   fn to_ffi(value: RustType) -> Self;
}

pub trait FromFFI<FFIType> {
   fn from_ffi(value: FFIType) -> Self;
}
```

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,203 @@
...
+	let _ = Box::from_raw(graph_state);
+}
+
+// Free GraphStates
+#[no_mangle]
+pub unsafe extern "C" fn free_graph_states() {
+	for graph_state in GRAPH_STATES.iter() {
+		let _ = Box::from_raw(*graph_state);
+	}
+}
```
JoeCap08055, line 183: If you free an individual graph, but don't remove it from the GRAPH_STATES vector, you're in danger of a double-free if you later call `free_graph_states()`

JoeCap08055: We also might want some kind of thread-safe vector for GRAPH_STATES; mutable static is not thread-safe.

JoeCap08055: ...and we might want to try and prevent a client double-free from calling `free_graph_state()` on the same pointer twice:
1. Check that pointer is in GRAPH_STATES vector before freeing
2. Take pointer from GRAPH_STATES vector
3. Free

JoeCap08055: ...or we could change GRAPH_STATES to a HashMap with integer keys, and instead of using pointers, the client just gets a graph ID to specify which graph...

JoeCap08055: :point_up: would still need to be thread-safe

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,216 @@
...
+
+// Collection of GraphStates
+static mut GRAPH_STATES: Option<Arc<Mutex<Vec<*mut GraphState>>>> = None;
+
+// Initialize GraphState
+#[no_mangle]
+pub unsafe extern "C" fn initialize_graph_state(
+	environment: *const Environment,
+) -> *mut GraphState {
+	let environment = &*environment;
```
aramikm, line 25: We need to wrap all these api calls in `catch_unwind`. Even though we are not going to have any expected panics but since having a panic is an undefined state we should handle them
https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/std/panic/fn.catch_unwind.html

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,216 @@
...
+	let rust_environment = environment_from_ffi(environment);
+	let graph_state = Box::into_raw(Box::new(GraphState::new(rust_environment)));
+
+	// Add state pointer to GRAPH_STATES vector if GRAPH_STATES is None
+	if GRAPH_STATES.is_none() {
+		GRAPH_STATES = Some(Arc::new(Mutex::new(Vec::new())));
```
aramikm, line 31: so if the state is none it is not going to add to the vector?

JoeCap08055 (commented):
Still not thread-safe:

* Modification of an existing graph is protected, but GRAPH_STATE can be set to `Some(...)` or `None` by multiple threads simultaneously.
* Suggest getting rid of `Option<...>` and just using `Arc<Mutex<Vec<GraphState>>>`.  Declare & initialize with an empty vector. When doing `free_graph_states()`, just use `Vec::clear()` && `Vec::shrink_to_fit()`

saraswatpuneet:
@JoeCap08055  in that case I will have to expose Graph types in bindings rational the Graph type itself is not threadsafe 
`user_key_manager: Rc<RefCell<dyn UserKeyManagerBase>>`

`Arc<Mutex<Vec<GraphState>>>` would require making this threadsafe deep down 

`c<RefCell<(dyn dsnp_graph_core::graph::key_manager::UserKeyManagerBase + 'static)>> cannot be sent between threads safely`
Lets revisit this

saraswatpuneet:
I believe in its current form , making it threadsafe  for GRAPH_STATES will require either
* Define C types for complete Graph
* Just do one instance of state per thread 
* Or just remove free GRAPH_STATES and keep shrinking the vector as states get removed

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,284 @@
...
+	if GRAPH_STATES.is_none() {
+		return 0
+	}
+	let graph_states = GRAPH_STATES.as_ref().unwrap().lock().unwrap();
+	graph_states.len()
+}
+
+// Get graph state from GRAPH_STATES at a given index
+#[no_mangle]
+pub unsafe extern "C" fn get_graph_state_at_index(index: usize) -> *mut GraphState {
```
JoeCap08055, line 96: This is why I suggested a HashMap instead of a Vec to store GraphState... index is kind of useless, as it may change as a result of an intervening call to `free_graph_state()`.

I would say either switch to a HashMap, or get rid of this function; we probably don't need it anyway, since we never return the index to the client. If we want to be able to iterate over all `GraphState`s, we should just provide access to an iterator.

core/src/graph/graph.rs
```diff
@@ -32,7 +30,7 @@ pub struct Graph {
 	user_id: DsnpUserId,
 	schema_id: SchemaId,
 	pages: PageMap,
-	user_key_manager: Rc<RefCell<dyn UserKeyManagerBase>>,
+	user_key_manager: Arc<RwLock<dyn UserKeyManagerBase + 'static + Send + Sync>>,
```
saraswatpuneet, line 33: Given `UserKeyManagerBase` is a trait, any implementation of this trait should be associated with `Send` and `Sync` to enforce thread safety

bridge/ffi/src/c_api.rs
```diff
@@ -245,22 +230,18 @@ pub unsafe extern "C" fn graph_get_one_sided_private_friendship_connections(
...
-	}
-	let mut graph_states = GRAPH_STATES.as_ref().unwrap().lock().unwrap();
-	if let Some(index) = graph_states.iter().position(|&state| state == graph_state) {
-		let _ = Box::from_raw(graph_states.remove(index));
-	}
+pub extern "C" fn free_graph_state(graph_state: *mut GraphState) {
+	let graph_state_box = unsafe { Box::from_raw(graph_state) };
+	let mut graph_states = GRAPH_STATES.lock().unwrap();
+	graph_states.retain(|x| !std::ptr::eq(x, &graph_state_box));
+	drop(graph_state_box);
```
JoeCap08055, line 227: This might be a double-deletion:
1. `graph_states.retain()` will drop the `GraphState` struct matching the pointer
2. Then we drop the pointer again explicitly.

Suggest:
* Eliminate `Box::from_raw()`; just use the pointer to call `graph_states.retain()` and the GraphState should be dropped.
* Might want to check that the graph was actually dropped (compare before & after count of `graph_states`... if this was the incorrect API instance, the graph would not be dropped; THEN we might want to drop it explicitly (or return error/false?)

saraswatpuneet: `graph_states.retain(|x| !std::ptr::eq(x.as_ref(), unsafe { &*graph_state })); `
drops it too as well from vector

bridge/ffi/src/c_api.rs
```diff
@@ -0,0 +1,254 @@
...
+pub unsafe extern "C" fn graph_contains_user(
+	graph_state: *mut GraphState,
+	user_id: *const DsnpUserId,
+) -> bool {
+	if graph_state.is_null() {
+		return false
+	}
+	let graph_state = &mut *graph_state;
+	let user_id = &*user_id;
+	graph_state.contains_user_graph(user_id)
```
aramikm, line 92: so I think the `catch_unwind` should be wrapping all api calls since you would not know which one might panic. You can implement something like the following to pass in a lambda function on all api calls
https://github.com/signalapp/libsignal/blob/0e2f87f39a2ebf8cc6e30fbff7fbf2c207b820da/rust/bridge/shared/src/ffi/mod.rs#L82-L95

core/src/api/api.rs
```diff
@@ -104,7 +102,7 @@ impl Transactional for GraphState {
 			}
 		}
 		self.user_map.commit();
-		self.shared_state_manager.deref().borrow_mut().commit();
+		self.shared_state_manager.write().unwrap().commit();
```
aramikm, line 105: hmmm, can this unwrap fail? What are the conditions?

bridge/ffi/src/c_api.rs
saraswatpuneet: Now that GraphState is thread safe, panic_unwind wont complain

https://github.com/LibertyDSNP/graph-sdk/pull/57/commits/27d24a2b6c5b0598a73ee017d6e3d3e6eb4e60f2

PR #58, opened 2023-05-09
added support for key rotation flow

### LINKED ISSUE #18: [GraphSDK] Key rotation and export process
## what
We would need to use added new key to encrypt and calculation PRI for the private graph pages

## Acceptance criteria
- Should be able to export a new key into data that can be written on chain
- After adding a new key `calculate_updates` should iterate through all pages for that user and re-encrypt and update PRids for all.

aramikm:
# Goal
The goal of this PR is to allow 2 things
- Export an itemized public key based on avro schema so users can publish their keys on chain
- Force calculate imported graph which would allow 3 main things
1. re-calculate pages using the latest key AKA key rotation
2. force refresh PRIDs using the latest keys
3. remove all empty pages

Closes #18 

# Discussion
- Based on the complexities and race conditions involved, decided to no allow users to publish a new key and rotate pages using in the same operation. Since we can not guarantee transactional update of chain. User first needs to publish the new key on chain and then in another operation it can use that key for encryption in pages.

# Checklist
- [x] Tests added


core/src/api/api.rs
```diff
@@ -180,6 +184,40 @@ impl GraphAPI for GraphState {
 		result
 	}
 
+	/// Export the graph pages for a certain user encrypted using the latest published key
+	fn force_recalculate_graphs(&self, user_id: &DsnpUserId) -> Result<Vec<Update>> {
```
aramikm, line 188: This function doesn't have a direct test since it would be an integration test and we have another ticket for it.

core/src/graph/shared_state_manager.rs
```diff
@@ -146,41 +151,33 @@ impl PublicKeyProvider for SharedStateManager {
...
 		Ok(result)
 	}
 
-	fn get_all_keys(&self, dsnp_user_id: DsnpUserId) -> Vec<&DsnpPublicKey> {
+	fn get_imported_keys(&self, dsnp_user_id: DsnpUserId) -> Vec<&DsnpPublicKey> {
 		let mut all_keys = vec![];
 		if let Some((v, _)) = self.dsnp_user_to_keys.get(&dsnp_user_id) {
 			all_keys.extend(&v[..]);
 		}
-		if let Some(k) = self.new_keys.get(&dsnp_user_id) {
```
aramikm, line 175: removing new keys from here since they can not being used until being publish on chain

core/src/api/api.rs
```diff
@@ -180,6 +184,40 @@ impl GraphAPI for GraphState {
...
+				.get_schema_id_from_connection_type(connection_type)
+				.ok_or(Error::msg(format!(
+					"Schema id for {:?} does not exist!",
+					connection_type
+				)))?;
+			let config = self
+				.get_dsnp_config(schema_id)
+				.ok_or(Error::msg("Dsnp config should is not supported!"))?;
+
+			let graph = user_graph.graph(&schema_id);
```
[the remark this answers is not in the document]
aramikm: moved into user_graph

core/src/api/api.rs
```diff
@@ -51,6 +51,10 @@ pub trait GraphAPI {
 	/// Apply Actions (Connect or Disconnect) to the list of pending actions for a users graph
 	fn apply_actions(&mut self, action: &[Action]) -> Result<()>;
 
+	/// Force re-calculates the imported graphs. This is useful to ensure the pages are using the
+	/// latest encryption key or refresh calculated PRIds or remove any empty pages and ...
+	fn force_recalculate_graphs(&self, user_id: &DsnpUserId) -> Result<Vec<Update>>;
+
```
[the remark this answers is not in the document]
aramikm: Would cause conflict when aggressive algorithm is implemented! keeping it separated for now.

core/src/api/api.rs
```diff
@@ -230,22 +234,14 @@ impl GraphAPI for GraphState {
...
-			None => return Err(Error::msg("user not present in graph state")),
+			None => return Err(Error::msg(format!("User {} graph is not imported!", user_id))),
 		};
 		let graph = user_graph.graph(&private_friendship_schema_id);
 		graph.get_one_sided_friendships()
 	}
 }
 
 impl GraphState {
-	fn get_dsnp_config(&self, schema_id: SchemaId) -> Option<DsnpVersionConfig> {
```
aramikm, line 241: moved into `UserGraph`

saraswatpuneet (approved):
👍

PR #66, opened 2023-05-12
Ffi clippy

saraswatpuneet:
# Goal
The goal of this PR is to address clippy failures


bridge/ffi/src/c_api.rs
```diff
@@ -262,37 +246,31 @@ pub unsafe extern "C" fn graph_get_one_sided_private_friendship_connections(
...
+// free graph state
 #[no_mangle]
 pub unsafe extern "C" fn free_graph_state(graph_state: *mut GraphState) {
 	let result = panic::catch_unwind(|| {
 		if graph_state.is_null() {
 			return
 		}
 		let mut graph_states = GRAPH_STATES.lock().unwrap();
-		graph_states.retain(|x| !std::ptr::eq(x.as_ref(), unsafe { &*graph_state }));
+		let index = graph_states.iter().position(|x| x as *const _ == graph_state).unwrap();
```
aramikm, line 260: This is interesting, I tried the following but it didn't work. I'm curious why it works with `position`

`graph_states.retain(|x| x as *const _ != graph_state);`

aramikm (approved):
Looks good to me!

bridge/ffi/src/c_api.rs
saraswatpuneet: Saw it later
Ideally eq should be used to compare addresses
But since we are storing the state itself and internally GraphState has eq derived we can safely compare the state and drop over using addresses

PR #71, opened 2023-05-15
feat: rework page-fullness approach and add aggressive algo

### LINKED ISSUE #61: [GraphSDK] Aggressive paginating algo
Blocked by #25 
Implement defined Algorithm to calculate and compute the results of changes for graphs.

## Additional data
Located at https://github.com/LibertyDSNP/frequency/blob/main/designdocs/graph_sdk.md

JoeCap08055:
# Goal
The goal of this PR is to add an "aggressive" algorithm to page update calculations that determines actual page fullness by performing compression & encryption.

Closes #61 

# Discussion

# Checklist
- [x] Tests added


core/src/graph/graph.rs
```diff
@@ -563,15 +615,82 @@ impl Graph {
...
+		}
+
+		if connection_type.privacy_type() == PrivacyType::Private &&
+			(dsnp_version_config.is_none() || keys.is_none())
+		{
+			return Err(Error::msg("Missing encryption key or config"))
+		}
+
+		let max_page_size = self.environment.get_config().max_graph_page_size_bytes as usize;
+		// Max page fullness leaves room for 2 uncompressed connections
```
aramikm, line 653: Would you mind add more details where this number 2 comes from?

core/src/graph/graph.rs
```diff
@@ -563,15 +615,82 @@ impl Graph {
...
+		let max_page_fullness = max_page_size -
+			(2 as usize *
+				(std::mem::size_of::<DsnpGraphEdge>() + std::mem::size_of::<DsnpPrid>()));
```
aramikm, line 656: There are cases that in memory size is different from serialized size due to padding and ... We would definitely need to check this with the actual serialized size.

core/src/graph/graph.rs
```diff
@@ -563,15 +615,82 @@ impl Graph {
...
+		if let Err(e) = temp_page.add_connection(connection_id) {
+			return Err(e)
 		}
 
-		todo!()
+		let page_blob = match connection_type {
+			ConnectionType::Follow(PrivacyType::Public) |
+			ConnectionType::Friendship(PrivacyType::Public) => temp_page.to_public_page_data(),
+			ConnectionType::Follow(PrivacyType::Private) => {
+				let keys = keys.as_ref().unwrap();
```
aramikm, line 667: unwraps should be eliminated and replaces with errors if possible.

core/src/graph/graph.rs
```diff
@@ -563,15 +615,82 @@ impl Graph {
...
+				let dsnp_version_config = dsnp_version_config.unwrap();
+				temp_page.clear_prids();
+				temp_page.to_private_page_data(dsnp_version_config, &keys)
+			},
+			ConnectionType::Friendship(PrivacyType::Private) => {
+				let keys = keys.as_ref().unwrap();
+				let dsnp_version_config = dsnp_version_config.unwrap();
+				self.apply_prids(&mut temp_page, &vec![], &keys)
```
aramikm, line 675:
```suggestion
				self.apply_prids(&mut temp_page, &vec![connection_id], &keys)
```

core/src/graph/graph.rs
```diff
@@ -563,15 +615,82 @@ impl Graph {
...
+					.expect("Error applying prids to page");
+				temp_page.to_private_page_data(dsnp_version_config, &keys)
+			},
+		};
+
+		match page_blob {
+			Ok(blob) =>
+				if blob.content.len() >= max_page_fullness {
```
aramikm, line 683: is this `>` or `>=` ?

core/src/graph/graph.rs
```diff
@@ -563,15 +615,82 @@ impl Graph {
...
+					Err(Error::msg("Page aggressively full"))
+				} else {
+					if let Err(e) = page.add_connection(connection_id) {
```
aramikm, line 686: what is the purpose of adding it again? when it's added on line 659?

core/src/graph/graph.rs
```diff
@@ -245,51 +245,107 @@ impl Graph {
...
+				while let Some(id_to_add) = add_iter.peek() {
+					if let Ok(_) = self.try_add_connection_to_page(
+						page,
+						id_to_add,
+						aggressive,
+						&encryption_key,
+						Some(dsnp_version_config),
+					) {
+						let _ = add_iter.next(); // TODO: prefer advance_by(1) once that stabilizes
+						continue
```
aramikm, line 260: this is not necessary and can be removed

core/src/graph/graph.rs
```diff
@@ -245,51 +245,107 @@ impl Graph {
...
+			for page in updated_pages.values_mut() {
+				while let Some(id_to_add) = add_iter.peek() {
+					if let Ok(_) = self.try_add_connection_to_page(
+						page,
+						id_to_add,
+						aggressive,
+						&encryption_key,
+						Some(dsnp_version_config),
+					) {
+						let _ = add_iter.next(); // TODO: prefer advance_by(1) once that stabilizes
```
aramikm, line 259: Would you mind explain more about todo?

core/src/graph/graph.rs
```diff
@@ -245,51 +245,107 @@ impl Graph {
...
+						&mut **page,
+						id_to_add,
+						true,
+						&encryption_key,
+						Some(dsnp_version_config),
+					)
+					.is_ok()
+				{
+					let _ = add_iter.next(); // TODO: prefer advance_by(1) once stabilized
+					if add_iter.peek().is_none() {
```
aramikm, line 334: do you really need this if you are doing the if Some check on line 345?

core/src/graph/graph.rs
```diff
@@ -245,51 +245,107 @@ impl Graph {
...
 						updated_pages.insert(page.page_id(), *page.clone());
-						current_page = None;
+						new_page = None;
 					}
-				},
-				None => return Err(Error::msg("Graph is full")), // todo: re-calculate updates with agressive fullness determination
+				} else {
+					new_page = None;
```
aramikm, line 339: what if the number of new added connections are more than 1 new page? I think maybe you would need to insert the previous page into updated pages and the try_add_connection_to_page should return more than success or failure, You need a third state which is failure due to page being full.

core/src/graph/graph.rs
```diff
@@ -1043,21 +1165,39 @@ mod test {
 				Rc::new(RefCell::from(SharedStateManager::new())),
 			))),
 		);
-		for p in page_builder.build() {
-			let _ = graph.create_page(&p.page_id(), Some(p)).expect("should create page!");
+
+		for mut p in page_builder.build() {
+			// let page_id = p.page_id();
```
aramikm, line 1170: commented

aramikm (commented):
Tremendous effort! Added some observations and potentially missing cases. I think we can refactor the code quite a bit since we are duplicating in a few places but that's not a blocker for this PR.

core/src/graph/graph.rs
JoeCap08055: Good point.

core/src/graph/graph.rs
JoeCap08055: It's just a cleaner way of advancing the iterator if we don't care about the return value. But it's currently unstable nightly.

core/src/graph/graph.rs
JoeCap08055: Right.

core/src/graph/graph.rs
```diff
@@ -620,33 +635,37 @@ impl Graph {
...
-	) -> Result<()> {
+	) -> Result<(), AddConnectionError> {
 		let connection_type = self.get_connection_type();
 		let max_connections_per_page =
 			*PAGE_CAPACITY_MAP.get(&connection_type).unwrap_or_else(|| {
 				let mut capacities: Vec<&usize> = PAGE_CAPACITY_MAP.values().collect();
 				capacities.sort();
 				capacities.first().unwrap() // default: return smallest capacity value
 			});
+		let encryption_key = match self.get_connection_type().privacy_type() {
```
aramikm, line 646: why duplicate it here if we already had access to it? It this for ease of testing?

core/src/graph/graph.rs
```diff
@@ -280,7 +300,6 @@ impl Graph {
 					&mut current_page,
 					id_to_add,
 					true,
-					&encryption_key,
```
aramikm: line 292-323 can be refactored like
```
for (_, page) in
			self.pages.inner().iter().filter(|(page_id, _)| !updated_keys.contains(page_id))
{
	let mut current_page = page.clone();
	let mut page_modified = false;
	while let Some(id_to_add) = add_iter.peek() {
		if let Ok(_) = self.try_add_connection_to_page(
			&mut current_page,
			id_to_add,
			true,
			Some(dsnp_version_config),
		) {
			page_modified = true;
			let _ = add_iter.next(); // TODO: prefer advance_by(1) once stabilized
		} else {
			// If we couldn't add a connection to the current page, it's full. Add this page
			// to the updated pages list and move on to the next page.
			break
		}
	}

	if page_modified {
		updated_pages.insert(current_page.page_id(), current_page.clone());
	}
	if let None = add_iter.peek() {
		break
	}
}
```

core/src/graph/graph.rs
```diff
@@ -620,33 +635,37 @@ impl Graph {
 		page: &mut GraphPage,
 		connection_id: &DsnpUserId,
 		aggressive: bool,
-		keys: &Option<ResolvedKeyPair>,
 		dsnp_version_config: Option<&DsnpVersionConfig>,
-	) -> Result<()> {
+	) -> Result<(), AddConnectionError> {
 		let connection_type = self.get_connection_type();
 		let max_connections_per_page =
```
aramikm, line 640: I think this should move inside` if !aggressive {`

core/src/graph/graph.rs
JoeCap08055: Yes; made it easier for testing. Now that I've got that all sorted, I could factor this out and pass in if we want.

core/src/graph/graph.rs
aramikm: it seems like we would never fill these pages to the max with this logic. the point of having some capacity inside pages was to allow insertions in future if that page got touched by a removal but based on current implementation it seems like this extra capacity will never be filled under any condition

core/src/graph/graph.rs
JoeCap08055: I'm not that concerned. Leaving space for 2 *uncompressed* connections is probably more than enough of a buffer, since (1) even worst case compression of a single connection won't double the size, and (2) we calculate page fullness using post-compression, post-encryption serialized data, so really the only reason to reserve space in the page would be to hedge against future on-chain modifications that introduce overhead to paginated storage.

I'm open to reducing the margin of error even further; but I think it's probably best to leave it as is for right now and revisit once we have real chain interactions to analyze.

core/src/graph/graph.rs
JoeCap08055: Added.

core/src/graph/graph.rs
JoeCap08055: Nice. Done.

core/src/graph/graph.rs
JoeCap08055: Not that important, since it only makes a difference of 1 byte, and since the size of the "safety padding" is larger than that, it's not a big deal.

But, you're right--it's `>`. Fixed. 😬

core/src/graph/graph.rs
JoeCap08055: Done.

core/src/graph/graph.rs
JoeCap08055: Removed extra space padding completely based on discussion.

core/src/graph/graph.rs
```diff
@@ -248,52 +249,95 @@ impl Graph {
...
-			if let Some((_page_id, page)) =
-				updated_pages.iter_mut().find(|(_, page)| !self.is_page_full(&page, false))
-			{
-				let id_to_add = add_iter.next().unwrap(); // safe to unwrap because we peeked above
-				page.add_connection(id_to_add)?;
-			} else {
+		let mut add_iter = ids_to_add.iter().cloned().peekable();
+		for aggressive in vec![false, true] {
+			for page in updated_pages.values_mut() {
+				self.add_to_page_until_full(page, &mut add_iter, aggressive, dsnp_version_config);
```
aramikm, line 255: nit: this would still iterate pages even though the `add_iter` might be finished

core/src/graph/graph.rs
```diff
@@ -569,15 +609,75 @@ impl Graph {
 	/// Determine if page is full
 	///  aggressive:false -> use a simple heuristic based on the number of connections
 	///  aggressive:true  -> do actual compression to determine resulting actual page size
-	pub fn is_page_full(&self, page: &GraphPage, aggressive: bool) -> bool {
+	pub fn try_add_connection_to_page(
```
aramikm, line 612: nit: Not a blocker but I think we should pass in the encryption key instead of fetching it here again. One way to facilitate tests for this might be to create a second method just for tests `#[cfg(test)]` (without encryption key) that gets the encryption key from `user_key_manager`

core/src/graph/graph.rs
```diff
@@ -569,15 +609,75 @@ impl Graph {
...
+
+		match page_blob {
+			Ok(blob) =>
+				if blob.content.len() > max_page_size {
+					Err(DsnpGraphError::PageAggressivelyFull)
+				} else {
+					if let Err(e) = page.add_connection(connection_id) {
+						return Err(e)
+					}
+
```
aramikm, line 676:
```suggestion
					page.add_connection(connection_id)?;

```

aramikm (approved):
Looks good! Thanks for refactoring.

saraswatpuneet (approved):

core/src/graph/graph.rs
JoeCap08055: Ah, right... now that we're returning real errors we can do that 👍🏻

core/src/graph/graph.rs
JoeCap08055: Done.

PR #72, opened 2023-05-15
api to expose imported public keys

### LINKED ISSUE #69: [GraphSDK] need new API endpoint to expose imported public key for a user
Since we are avro encoding public keys inside chain for any wallet to import related private keys they need to know what public keys are associated for any user. In that regard we would need a new endpoint that would return imported public keys for a user

aramikm:
# Goal
The goal of this PR is provide an api to expose imported public keys

Closes #69 

# Checklist
- [x] FFI binding added 
- [x] Tests added

### DIFF
--- a/bridge/ffi/src/c_api.rs
+++ b/bridge/ffi/src/c_api.rs
@@ -249,6 +249,27 @@ pub unsafe extern "C" fn graph_get_one_sided_private_friendship_connections(
 	result.unwrap_or(GraphConnections { connections: std::ptr::null_mut(), connections_len: 0 })
 }
 
+// Get a list published and imported public keys associated with a user
+#[no_mangle]
+pub unsafe extern "C" fn graph_get_public_keys(
+	graph_state: *mut GraphState,
+	user_id: *const DsnpUserId,
+) -> DsnpPublicKeys {
+	let result = panic::catch_unwind(|| {
+		if graph_state.is_null() {
+			return DsnpPublicKeys { keys: std::ptr::null_mut(), keys_len: 0 }
+		}
+		let graph_state = &mut *graph_state;
+		let user_id = &*user_id;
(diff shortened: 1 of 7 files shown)


saraswatpuneet (approved):
Look good! Want to add the same to FFI?

aramikm:
Oops forgot about that. Good point. I'll add

saraswatpuneet (approved):

bridge/ffi/src/c_api.rs
```diff
@@ -249,6 +249,27 @@ pub unsafe extern "C" fn graph_get_one_sided_private_friendship_connections(
...
+	let result = panic::catch_unwind(|| {
+		if graph_state.is_null() {
+			return DsnpPublicKeys { keys: std::ptr::null_mut(), keys_len: 0 }
+		}
+		let graph_state = &mut *graph_state;
+		let user_id = &*user_id;
+		let keys = graph_state.get_public_keys(user_id).unwrap_or_default();
+		let ffi_keys = dsnp_public_keys_to_ffi(keys);
+		let keys_len = ffi_keys.len();
+		let keys_ptr = ManuallyDrop::new(ffi_keys).as_mut_ptr();
```
JoeCap08055, line 267: Don't we also need a method to manually release this memory at some point?

aramikm: We should, should we do that in this layer or that needs to be done from the layer that consumes FFI?

JoeCap08055: The layer that consumes the FFI will need to invoke it, but we still need to provide a native Rust method to release the memory. For instance, C or Java code can't natively de-allocate memory that was allocated in Rust. So I think we will need a `free_public_keys()` function.

aramikm: I'll take care of it in a future PR in that case.

PR #73, opened 2023-05-16
import keys via connect action

### LINKED ISSUE #70: [GraphSDK] allow to import public keys associated with a user in Connect API
For Private friendship users we would need to import all the keys associated with that new connection. This would force us to do multiple api calls to fulfill all requirements of a simple **Connect** . To simply the connect process it would be preferable to allow importing keys with **Connect** action so there are less back and forth steps are required.

aramikm:
# Goal
The goal of this PR is make apply_actions api self contained so we can import keys while adding a new connection.
This is necessary to calculate PRIds for private friendship connections

Closes #70 

# Checklist
- [x] Tests added
- [x] FFI updated

### DIFF
--- a/bridge/ffi/src/bindings.rs
+++ b/bridge/ffi/src/bindings.rs
@@ -210,6 +210,9 @@ pub enum Action {
 
 		/// connection details
 		connection: Connection,
+
+		/// optional key to import
+		dsnp_keys: *mut DsnpKeys,
 	},
 
 	/// an action that defines removing an existing connection from social graph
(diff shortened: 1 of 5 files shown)


bridge/ffi/src/bindings.rs
```diff
@@ -210,6 +210,9 @@ pub enum Action {
 
 		/// connection details
 		connection: Connection,
+
+		/// optional key to import
+		dsnp_keys: *mut DsnpKeys,
```
aramikm, line 215: I'm not sure about this. Trying to represent a ffi compatible `Option<DsnpKeys>`

PR #74, opened 2023-05-16
Better errors

### LINKED ISSUE #28: [GraphSDK] better errors

saraswatpuneet:
# Goal
The goal of this PR is to define custom graph sdk type errors and modify underlying traits and implementation to return custom type instead

Closes #28 

# Discussion

# Checklist
- [ ] Tests added


config/src/errors.rs
```diff
@@ -0,0 +1,180 @@
+//! Errors for graph-sdk crate
```
saraswatpuneet, line 1: Various errors returned from sdk

config/src/errors.rs
```diff
@@ -0,0 +1,180 @@
...
+//!
+
+use thiserror::Error;
+
+pub type DsnpGraphResult<T> = std::result::Result<T, DsnpGraphError>;
```
saraswatpuneet, line 6: graph sdk `Result` type

config/src/errors.rs
```diff
@@ -0,0 +1,180 @@
...
+	// Invalid Page ID
+	#[error("Invalid Page ID: {0}")]
+	InvalidPageId(String),
+
+	// Page error
+	#[error("Page error: {0}")]
+	PageError(String),
+
+	// "Connection from {} to {} does not exists to be disconnected!"
+	#[error("Connection from {0} to {1} does not exists!")]
```
JoeCap08055, line 67: s/exists/exist/

config/src/errors.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+impl From<apache_avro::Error> for DsnpGraphError {
+	fn from(error: apache_avro::Error) -> Self {
+		// Convert the underlying error to DsnpGraphError variant
+		// based on your error handling logic
+		// For example:
+		DsnpGraphError::AvroError(error.to_string())
+	}
+}
```
JoeCap08055, line 180: Why not instead define `DsnpGraphError::AvroError` as:
```
#[error(transparent)]
AvroError(#[from] apache_avro::Error)
```

...will auto-generate a `From<apache_avro::Error>` implementation

config/src/errors.rs
```diff
@@ -0,0 +1,180 @@
+//! Errors for graph-sdk crate
+//!
+
+use thiserror::Error;
+
+pub type DsnpGraphResult<T> = std::result::Result<T, DsnpGraphError>;
+
+#[derive(Debug, Error)]
+pub enum DsnpGraphError {
```
JoeCap08055, line 9: q: why are all values `String` instead of, for instance, `DsnpUserId` or other types? Is it specifically for C-compatibility? It seems like the `thiserror` crate gives us a lot of flexibility to keep the underlying data types while still producing a `Display` trait with a formatted string.

aramikm (commented):
Great work! Added some suggestions and also I think we would need to modify the tests to check against those specific errors instead of currently only checking `is_err`

config/src/errors.rs
JoeCap08055, line 9: Do we want to `#[repr(C, u8)]` this? Or otherwise we'll need to wrap it in something for the FFI.

config/src/errors.rs
saraswatpuneet: Yep i like that

config/src/errors.rs
```diff
@@ -0,0 +1,180 @@
...
+	// Errors related to Avro
+	#[error("Avro error: {0}")]
+	AvroError(String),
```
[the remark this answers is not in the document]
saraswatpuneet: I sorted them but since they are not unit variants and contains tuples we cannot achieve pinning the values like =1 etc.
One way to achieve this especially for ffi is assign them some code via function call, 

Let me know how you feel about it

Makefile
```diff
@@ -15,9 +15,10 @@ clippy:
 	@echo "Running Cargo clippy..."
 	@cargo clippy --all --all-features --all-targets -- -D warnings
 
-.PHONY: fmt
-fmt:
+.PHONY: format
+format:
 	@echo "Running Cargo fmt..."
+	@cargo fmt
```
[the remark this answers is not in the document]
saraswatpuneet: I meant to separate the formatting and checking, but yes in CI we should only check than format

core/src/dsnp/schema.rs
```diff
@@ -87,10 +88,11 @@ mod test {
 		let deserialized = SchemaHandler::read_public_key(&serialized);
 
 		assert!(deserialized.is_err());
-		assert!(matches!(
-			deserialized.unwrap_err().downcast_ref::<AvroError>(),
-			Some(AvroError::ConvertI64ToUsize(_, _))
-		))
+		let returned_err = deserialized.unwrap_err();
+		match returned_err {
```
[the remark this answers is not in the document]
saraswatpuneet: Yep, that suggest from Joe help to match exact error

core/src/dsnp/api_types.rs
```diff
@@ -1,8 +1,14 @@
 use crate::dsnp::{dsnp_configs::KeyPairType, dsnp_types::DsnpUserId};
-pub use dsnp_graph_config::{ConnectionType, PrivacyType};
+pub use dsnp_graph_config::{ConnectionType, PageId as PageIdType, PrivacyType};
 use dsnp_graph_config::{GraphKeyType, SchemaId};
 use std::{cmp::Ordering, fmt::Debug};
 
+/// Graph page id
+pub type PageId = PageIdType;
+
```
JoeCap08055, line 8: I think if you just do `pub use dsnp_graph_config::PageId` it will re-export; you don't need to redefine the `PageId` type again.

core/src/dsnp/dsnp_types.rs
```diff
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use dsnp_graph_config::DsnpUserId as DsnpUserIdType;
```
JoeCap08055, line 2: Same as above comment on `PageId`

aramikm (approved):
Looks great!

config/src/errors.rs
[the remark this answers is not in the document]
JoeCap08055: Not necessarily true; for instance, given the following enum:
```
#[repr(C)]
pub enum MyError {
   Error1 = 1,
   Error2(u8) = 2,
   Error3(String) = 3,
   Error4 { uid: u8, name: String } = 4,
}
```

cbindgen does generate integer-valued C-style enums. Here is the cbindgen output for that enum:
```
typedef struct String String;

typedef enum {
  Error1 = 1,
  Error2 = 2,
  Error3 = 3,
  Error4 = 4,
} MyError_Tag;

typedef struct {
  uint8_t uid;
  String name;
} Error4_Body;

typedef struct {
  MyError_Tag tag;
  union {
    struct {
      uint8_t error2;
    };
    struct {
      String error3;
    };
    Error4_Body error4;
  };
} MyError;
```

See here for more explanation:
https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#:~:text=Explicit%20repr%20annotation%20with%20C%20compatibility

It may not do any good in Rust, but the error struct exposed to the C API would expose the integer value of the enum.

PR #76, opened 2023-05-19
init: java/jni project

### LINKED ISSUE #63: [GraphSDK] Setup the structure of Java Project and Jni Bridge for SDK
Setup the bare minimum of Java classes and libraries that calls Rust code and Jni bridge in Rust

We are choosing to use **jni** crate to ensure memory safety and better integration with JVM and there is an example of that in https://github.com/jni-rs/jni-rs/tree/master/example that can be used to setup the details.

for more sophisticated setup can look at https://github.com/signalapp/libsignal/tree/main

aramikm:
# Goal
The goal of this PR is to setup the folders and project for java and jni

Closes #63 

# Discussion
- Currently we are not using `jni/cbindgen.toml` directly. The reason is that we are exposing struct and types and enums in our api that is not easily translatable to java. The way Signal works is that they only expose functions and they have a python script that uses `bindgen` output and generates a `Native.java`. In that way you can only change the `Rust` side and it would generate the compatible the native api signature in `java`. The catch is that it would be harder to expose `Structs` and `enums` and non primitive types in the api and you would need to serialize and deserialize to move everything between `java` and `rust` that is not a primitive type. I'm looking into `JNI` more deeply to understand what are the best practices and if it does make sense for us to do something like that or not. 
- Currently the example `hello` jni is being called from java and executed in rust so the connection is established.
- gradle related files are all auto-generated, we might need to modify them in future but it looks fine for now

# Checklist
- [x] Tests added for jni connection


saraswatpuneet (approved):
Nice scaffolding

bridge/jni/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "dsnp-graph-sdk-jni"
```
JoeCap08055, line 2: nit: Let's standardize our crate naming; currently we have some `dsnp-graph-*` and some `dsnp-graph-sdk-*`. My preference is for `dsnp-graph-*` (without SDK), but as long as we're consistent.

bridge/jni/src/api.rs
```diff
@@ -0,0 +1,29 @@
+use jni::{
+	objects::{JClass, JObject, JString},
+	JNIEnv,
+};
+
+#[no_mangle]
+pub extern "system" fn Java_frequency_Native_hello<'local>(
```
JoeCap08055, line 7: We probably don't need `extern "system"` here--that's more for interacting with OS libraries and system calls. For exposing Rust functions to other languages via a C-compatible ABI, just `extern "C"` is the way.

java/gradle/wrapper/gradle-wrapper.jar
JoeCap08055, line 1: Does this need to be committed?

java/lib/src/main/java/frequency/Native.java
JoeCap08055, line 1: Hmm, looks familiar... 😅

bridge/jni/Cargo.toml
aramikm: let's talk about this in next graph-sdk meeting

bridge/jni/src/api.rs
aramikm: Thius is the way javac -h generates the jni functions and even if you look at their example you see the same https://github.com/jni-rs/jni-rs/blob/master/example/mylib/src/lib.rs#L33. I think both of them work but I'm not sure which way is better.

java/gradle/wrapper/gradle-wrapper.jar
aramikm: yes

java/lib/src/main/java/frequency/Native.java
aramikm: Lol, yeah shameless borrowing. let me know if you have a better name.

bridge/jni/src/api.rs
JoeCap08055: I think that will break on Win32. Everything I've looked up says JNI expects a "C" ABI. "system" maps to "stdcall" on Win32. We would only want that if calling WIn32 system libraries from Rust code.

On any platform *but* Win32, `extern "C"` == `extern "system"`

JoeCap08055: I think you would only use "system/stdcall" if you were:
* declaring an external Win32-native DLL function for Rust to be able to call
* writing a native Win32 Rust library that other native Win32 DLLs or apps (not JNI) will call using the platform native ABI

PR #80, opened 2023-05-22
FFI: Result type and safe handling of pointers

### LINKED ISSUE #78: [GraphSDK] integration of better errors with FFI
# Problem
We would like to have a more in depth and better support for errors in FFI bridge.

saraswatpuneet:
# Goal
The goal of this PR is 

Closes #78 

# Discussion

# Checklist
- [x] Tests added


bridge/ffi/src/errors.rs
```diff
@@ -0,0 +1,29 @@
+use dsnp_graph_config::errors::DsnpGraphError;
+
+// Opaque handle for DsnpGraphError
+pub struct DsnpGraphErrorHandle {
+	// to make errors opaque this field must be private
+	// it enables to create a pointer to the error
+	error: *mut DsnpGraphError,
```
saraswatpuneet, line 7: we can expose underlying type transparently due to error types, so keeping it opaque as consumer would call provided functions to get error code and message

bridge/ffi/src/lib.rs
```diff
@@ -9,12 +9,42 @@
...
 #[cfg(test)]
 mod tests;
+
+use std::ptr::NonNull;
+
+#[repr(C)]
+#[derive(Debug, Clone, PartialEq)]
+pub struct DsnpGraphFFIResult<T, E> {
+	pub result: Option<NonNull<T>>,
+	pub error: Option<NonNull<E>>,
```
aramikm, line 31: I was imagining that maybe C enum Error might be sufficient. Any reason why we need generic (dynamic) errors?

bridge/ffi/src/lib.rs
```diff
@@ -9,12 +9,42 @@
...
 
 #[cfg(test)]
 mod tests;
+
+use std::ptr::NonNull;
+
+#[repr(C)]
+#[derive(Debug, Clone, PartialEq)]
+pub struct DsnpGraphFFIResult<T, E> {
+	pub result: Option<NonNull<T>>,
```
aramikm, line 30: What are the advantages of `NonNull` ?

aramikm (commented):
Great work! I added some questions due to missing some context.

bridge/ffi/src/lib.rs
saraswatpuneet: few that i learnt
* NonNull can be used to represent *mut/ *const types 
* Provides a safe abstraction for raw pointer (its essentially is kind of a smart pointer)
* NonNull types ensure the underlying pointer is never a null pointer and wrapping it around Option gives us ability to represent `NULL` type on c side while also being safe of rust side
* In general safer and sound in comparison to putting raw pointers there

bridge/ffi/src/lib.rs
saraswatpuneet: Just a design pattern so any changes to error types wont warrant changing this generic type always, but yes can be pinned to exact error type too, either or works

bridge/ffi/src/c_api.rs
```diff
@@ -232,54 +360,86 @@ pub unsafe extern "C" fn graph_get_connections_without_keys(
...
 	let result = panic::catch_unwind(|| {
 		if graph_state.is_null() {
 			return
 		}
 		let mut graph_states = GRAPH_STATES.lock().unwrap();
-		let index = graph_states.iter().position(|x| x as *const _ == graph_state).unwrap();
-		graph_states.remove(index);
+		let index =
+			graph_states.iter().position(|x| x.as_ref() as *const _ == graph_state).unwrap();
+		let _ = ManuallyDrop::new(graph_states.remove(index));
```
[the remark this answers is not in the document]
saraswatpuneet: Yes it should, but i wanted it to be more predictable and wanting to drop in the code exclusively then depending on vec  to do it for us, thought this is cleaner and easy to understand given this is where we want to be very sure given its the state.

do you prefer otherwise?

bridge/ffi/src/c_api.rs
```diff
@@ -331,3 +491,34 @@ pub unsafe extern "C" fn free_graph_dsnp_public_keys(public_keys: *mut DsnpPubli
 	});
 	result.unwrap_or(())
 }
+
+// Free DsnpGraphErrorHandle
+#[no_mangle]
+pub extern "C" fn free_dsnp_graph_error(error: *mut DsnpGraphErrorHandle) {
+	if !error.is_null() {
+		unsafe {
+			let _ = Box::from_raw(error);
```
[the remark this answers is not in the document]
saraswatpuneet: function `dsnp_graph_error_code` just returns a numeric code that we have set in `DsnpGraphError`  not sure if i understood the question

bridge/ffi/src/c_api.rs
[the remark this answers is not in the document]
aramikm: I'm just worried about double-free than anything. I'm more leaning towards depending on Rust for managing the memory as much as possible and if there are cases that we must handle it manually then it's fine

bridge/ffi/src/c_api.rs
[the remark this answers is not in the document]
aramikm: I I see the different use is based on the naming, my mistake!

bridge/ffi/src/c_api.rs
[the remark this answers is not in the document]
saraswatpuneet: Yeah no worries just pushed an update, letting rust do its thingy

bridge/ffi/src/c_api.rs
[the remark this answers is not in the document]
saraswatpuneet: good reminder, need to panic unwind these new functions

bridge/ffi/src/c_api.rs
```diff
@@ -331,3 +457,34 @@ pub unsafe extern "C" fn free_graph_dsnp_public_keys(public_keys: *mut DsnpPubli
...
+// Get error message
+#[no_mangle]
+pub unsafe extern "C" fn dsnp_graph_error_message(error: *const GraphError) -> *const c_char {
+	if let Some(error) = unsafe { error.as_ref() } {
+		let error_msg = CString::new(error.error_message()).unwrap_or_default();
+		error_msg.into_raw()
+	} else {
+		std::ptr::null()
+	}
+}
```
JoeCap08055, line 490: Don't we also need a "free" method for the returned error string?

bridge/ffi/src/lib.rs
JoeCap08055: It seems like only one of `result`, `error` will ever be `Some(value)`, while the other would be `None`. If that's the case, would it make sense for this to be a union instead of a struct?

bridge/ffi/src/lib.rs
```diff
@@ -9,12 +9,42 @@
...
+pub use errors::*;
 
 #[cfg(test)]
 mod tests;
+
+use std::ptr::NonNull;
+
+#[repr(C)]
+#[derive(Debug, Clone, PartialEq)]
+pub struct FFIResult<T, E> {
```
JoeCap08055, line 29: It seems like only one of result, error will ever be Some(value), while the other would be None. If that's the case, would it make sense for this to be a union instead of a struct?

bridge/ffi/src/c_example/main.c
```diff
@@ -12,66 +13,72 @@
...
-    free_graph_state(graph_state);
+    ImportBundle importbundle;
+    // Set the values of the importbundle struct
+    // ...
 
+   FFIResultboolGraphError importresult = graph_import_users_data(graphstate, &importbundle, 1);
+    ASSERT(importresult.error != NULL, "Expected error to import users data");
+    // get the error message
+    const char* errormessage = dsnp_graph_error_message(importresult.error);
+    ASSERT(errormessage != NULL, "Failed to get error message");
```
JoeCap08055, line 44: Wouldn't we also need to free the returned error message?

bridge/ffi/src/lib.rs
saraswatpuneet: i like the suggestion but then we cant derive fancy things on unions like `Debug` etc? especially debug trait is ideal here

Open to changing it if preferred

saraswatpuneet: Just realized it will be warranting me to change a lot around so we should do it as an enhancement

aramikm (approved):
This is great! Good to go!

PR #84, opened 2023-05-23
feat(#64): initial development of rust/jni logging infrastructure

### LINKED ISSUE #64: [GraphSDK] implement Log handling between jni and java code
1- Check to see what happens when we write logs directly in rust sdk and if it is shown in Java logs
2- Setup the logs call between jni and java to be able to call java log function from rust
3- Setup different logging levels

Related to #43

JoeCap08055:
# Goal
The goal of this PR is to develop a logging infrastructure to propagate logging from Rust to a JVM-based logger.

Closes #64 

# Checklist
- [ ] Tests added


bridge/jni/src/logger.rs
```diff
@@ -0,0 +1,195 @@
...
+	let level: SLF4JLogLevel = match max_level.try_into() {
+		Ok(level) => level,
+		_ => panic!("invalid log level"),
+	};
+
+	log::set_max_level(log::Level::from(level).to_level_filter())
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn Java_org_signal_libsignal_internal_Native_Logger_1Initialize(
```
aramikm, line 150:
```suggestion
pub unsafe extern "C" fn Java_io_amplica_graphsdk_Native_Logger_1Initialize(
```

bridge/jni/src/logger.rs
```diff
@@ -0,0 +1,195 @@
...
+					"Duplicate logger initialization ignored for {}",
+					env!("CARGO_PKG_NAME")
+				);
+			},
+		}
+	});
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn Java_org_signal_libsignal_internal_Native_Logger_1SetMaxLevel(
```
aramikm, line 189:
```suggestion
pub unsafe extern "C" fn Java_io_amplica_graphsdk_Native_Logger_1SetMaxLevel(
```

bridge/jni/src/logger.rs
aramikm: the above is the latest packaging but if you want it to work with your current java packages this should be 
`Java_frequency_Native_Logger_1Initialize(`

scripts/build_jni.sh
```diff
@@ -7,11 +7,12 @@ JNI_NAME=dsnp_graph_sdk_jni
 
 cargo build -p dsnp-graph-sdk-jni --release
 
+install -d ${BACKEND_LIB_DIR}
```
aramikm, line 10: nice!

aramikm (approved):
Looks great!

bridge/jni/src/logger.rs
```diff
@@ -0,0 +1,223 @@
...
+		let throwable = env.exception_occurred()?;
+		if **throwable == *JObject::null() {
+			match result {
+				Ok(r) => Ok(r),
+				Err(e) => {
+					eprintln!("Error logging to JVM: {:?}\n{}", e, message);
+					Err(e)
+				},
+			}?;
+		// result?;
```
aramikm, line 112: is this leftover?

bridge/jni/src/logger.rs
```diff
@@ -0,0 +1,223 @@
...
+				let backtrace_mode = {
+					cfg_if::cfg_if! {
+						if #[cfg(target_os = "android")] {
+							log_panics::BacktraceMode::Unresolved
+						} else {
+							log_panics::BacktraceMode::Resolved
+						}
+					}
+				};
+				log_panics::Config::new().backtrace_mode(backtrace_mode).install_panic_hook();
```
aramikm, line 179: is there a way to test panic hook?

java/frequency_Native.h
```diff
@@ -0,0 +1,37 @@
+/* DO NOT EDIT THIS FILE - it is machine generated */
```
aramikm, line 1: I don't think should get committed.

java/lib/build.gradle
```diff
@@ -25,6 +25,11 @@ dependencies {
 
     // This dependency is used internally, and not exposed to consumers on their own compile classpath.
     implementation 'com.google.guava:guava:31.1-jre'
+
+    implementation 'org.slf4j:slf4j-api:2.0.7'
+    // implementation 'org.slf4j:slf4j-simple:2.0.7'
```
aramikm, line 30: leftover?

java/lib/hs_err_pid92538.log
```diff
@@ -0,0 +1,1552 @@
+#
```
aramikm, line 1: should not get commited!

java/lib/src/test/java/frequency/LibraryTest.java
```diff
@@ -6,13 +6,133 @@
...
+        log(Level.INFO.toInt(), "This is an info log");
+        log(Level.WARN.toInt(), "This is a warning log");
+        log(Level.ERROR.toInt(), "This is an error log");
+        assertEquals(false, testLogsForPattern(Level.DEBUG, "This is a debug log"));
+        assertEquals(false, testLogsForPattern(Level.INFO, "This is an info log"));
+        assertEquals(false, testLogsForPattern(Level.WARN, "This is a warning log"));
+        assertEquals(true, testLogsForPattern(Level.ERROR, "This is an error log"));
+    }
+
+    private static native void log(int level, String message);
```
aramikm, line 137: Should we keep all native functions inside Native even if it's for tests?

aramikm (commented):
Looks great, It seems like you have some leftover files.

java/lib/src/test/java/frequency/LibraryTest.java
JoeCap08055: I initially tried that, but Java doesn't have conditional compilation, and I didn't want to pollute the lib/jar in production code with functions that are only for test. I already had to do that on the Rust side, but I don't want to propagate that. At least the Rust function is technically valid for production (though unnecessary)

java/lib/build.gradle
JoeCap08055: Fixed.

bridge/jni/src/logger.rs
```diff
@@ -0,0 +1,223 @@
...
+	logger_class: GlobalRef,
+}
+
+impl SLF4JLogger {
+	fn new(env: JNIEnv, logger_obj: JObject) -> jni::errors::Result<Self> {
+		Ok(Self { vm: env.get_java_vm()?, logger_class: env.new_global_ref(logger_obj)? })
+	}
+
+	fn log_impl(&self, record: &log::Record) -> jni::errors::Result<()> {
+		let mut env = self.vm.attach_current_thread()?;
```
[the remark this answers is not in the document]
JoeCap08055: Good catch. Captured from Signal, but we don't need it, since all of our code is initiated through the Java side, and the native side doesn't have any independent threads.

bridge/jni/src/logger.rs
JoeCap08055: I'll look into it.

PR #85, opened 2023-05-23
sdk protobuf support

### LINKED ISSUE #81: [GraphSDK] Define protobuf types for API inputs and Outputs
# Problem
The details of why we decided to go this route is mentioned in #77

# Acceptence criterea
- Define protobuf types for api input and outputs
- Generate Rust code for defined protobuf types
- Generate Java code for defined protobuf types

aramikm:
# Goal
The goal of this PR is to add serialization and deserialization support between Java and Rust using protobuf

Closes #81 

# Discussion
- Added protobuf code generation support for Rust and java using `protoc`
- Created mapping for `initializeGraphState` api
- All Java files inside `java/lib/src/main/java/io/amplica/graphsdk/models/` are generated
- All Rust files inside `bridge/common/src/proto_types/` are generated

# Checklist
- [x] Java tests cover end to end serialization and deserialization


bridge/common/protos/input.proto
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    oneof inner {
+      ConnectAction connect_action = 1;
+      DisconnectAction disconnect_action = 2;
+      AddGraphKey add_key_action = 3;
+    }
+  }
+  repeated Action actions = 1;
+}
```
saraswatpuneet, line 75: are there some crates like https://docs.rs/autoproto/latest/autoproto/
that can automate this?

saraswatpuneet (approved):
looks good to me! nothing much to add just one comment

bridge/common/protos/input.proto
aramikm: I think this is dependent on `prost` crate which we are not using and it says one of it's limitations is exactly what we are doing `nested oneof ` to support Rust style enums

JoeCap08055:
Maybe we should integrate building the generated proto code into the build process, instead of checking the generated files in?

aramikm:
I think we will eventually use `build.rs` to generate rust and java files when `protos` changed but for now to be able to run the tests we need those classes and structs to check-in.

Issue #83, opened 2023-05-22
[GraphSDK] Implement logic inside of defined JNI apis


aramikm:
Since the apis are defined by #82 we will need to implement the type mappings and logic inside JNI functions.
Track memory pointers and make sure the memory passed to be release is already allocated in Rust

PR #87, opened 2023-05-24
define jni apis

### LINKED ISSUE #82: [GraphSDK] Define jni apis
# Problem
We need to be able to call all defined api's from java to Rust

# Acceptence Criterea
- Define all apis and (destructors if there are any memory allocation in the Rust) in Java side
- Define all apis and (destructors if there are any memory allocation in the Rust) in JNI side.

# Note
The type mapping and logic implementation has it's own ticket so this is just the methods without any implementation.

aramikm:
# Goal
The goal of this PR is to define all jni apis in rust and java side

Closes #82 

# Discussion
- Defines apis in jni side
- Define apis in Java side
- Some Java utility implementations
- since this is only the definition no tests are added. Tests will be added after complete implementation

### DIFF
--- a/bridge/jni/src/api.rs
+++ b/bridge/jni/src/api.rs
@@ -2,7 +2,7 @@ use crate::mappings::map_to_environment;
 use dsnp_graph_core::api::api::GraphState;
 use jni::{
 	objects::{JByteArray, JClass, JObject, JString},
-	sys::jlong,
+	sys::{jboolean, jint, jlong},
 	JNIEnv,
 };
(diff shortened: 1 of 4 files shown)


saraswatpuneet (approved):
Lgtm: Assuming these are not going to change in terms of their definition and implementation will come soon

aramikm:
> Lgtm: Assuming these are not going to change in terms of their definition and implementation will come soon

The implementation will follow in another PR related to #83

PR #89, opened 2023-05-26
implemented jni apis

### LINKED ISSUE #83: [GraphSDK] Implement logic inside of defined JNI apis

aramikm:
# Goal
The goal of this PR is implemented rest of JNI apis

Closes #83 

# Discussion
- implements mapping between Jni, Proto and Rust types
- Some minor cleanup in Java side for better usability
- Merging this would feature complete the SDK but there are some leftovers like integrating exception handling in JNI and ...
- Memory allocation and removal seems to be working as expected but will need to do more memory specific testing in future to make sure
# Checklist
- [x] All apis are tested and works as expected. Only `oneSidedFriendship` is not tested since it takes more time to setup the test and will be done in future PR (not a blocker to close this)

### DIFF
--- a/bridge/jni/src/api.rs
+++ b/bridge/jni/src/api.rs
@@ -1,11 +1,19 @@
-use crate::mappings::map_to_environment;
-use dsnp_graph_core::api::api::GraphState;
+use crate::mappings::{
+	convert_jboolean, map_to_actions, map_to_environment, map_to_imports, serialize_config,
+	serialize_dsnp_users, serialize_graph_edges, serialize_graph_updates, serialize_public_keys,
+};
+use dsnp_graph_config::{DsnpUserId, SchemaId};
+use dsnp_graph_core::api::api::{GraphAPI, GraphState};
 use jni::{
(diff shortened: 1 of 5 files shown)


saraswatpuneet (commented):
not sure if it is missed, is this a double free?

https://github.com/LibertyDSNP/graph-sdk/blob/9aedf23d50b6036dbfe4dc89d04373195a923010/bridge/jni/src/api.rs#L80

aramikm:
> not sure if it is missed, is this a double free?
> 
> https://github.com/LibertyDSNP/graph-sdk/blob/9aedf23d50b6036dbfe4dc89d04373195a923010/bridge/jni/src/api.rs#L80

That is intentional and actually the line that **frees** the memory. The removal from `GRAPH_STATES` Vec is just removing a `jlong` which even though is the address of the state in memory but it's completely dissociated from compilers perspective.

PR #90, opened 2023-05-31
GraphSDk -> move to kotlin

### LINKED ISSUE #68: Package publish
As dev I would like to get access to the graphsdk package.

An example from Peter's team
https://github.com/AmplicaLabs/saas-frequency-client/blob/main/build.gradle.kts

saraswatpuneet:
# Goal
The goal of this PR is to release jar for graphsdk
Closes #68


saraswatpuneet:
Published jar from job

https://github.com/orgs/LibertyDSNP/packages?repo_name=graph-sdk

saraswatpuneet:
Have few questions
* Where else are we publishing ?
* Naming convention

.github/workflows/pr-verify.yml
```diff
@@ -161,3 +161,31 @@ jobs:
...
+          java-version: '17'
+          distribution: 'adopt'
+      - name: Build JNI package
+        run: make build-jni
+      - name: Build and Publish with Gradle
+        working-directory: java
+        run: ./gradlew build
+      - name: Publish to GitHub Packages
+        working-directory: java
+        run: ./gradlew publish
```
aramikm, line 186: should we keep the build and publish under the same job or separate it to allow more manual publish?

java/lib/build.gradle.kts
```diff
@@ -0,0 +1,77 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    // Apply the kotlin-jvm plugin for Kotlin JVM projects.
+    kotlin("jvm") version "1.6.21"
+    // Apply the maven-publish plugin for publishing the library.
+    id("maven-publish")
+}
+
+group = "io.amplica.frequency.graph-sdk"
```
aramikm, line 10: I think our group would be `io.amplica.graphsdk` based on our current java packages.

aramikm (approved):
I think it uses the artifact name as the suffix like for us since it's `lib` it create `.lib` but it doesn't seem to matter

Looks good , I added two comments

aramikm (commented):
I think `settings.gradle` should also renamed to `settings.gradle.kts`

.github/workflows/pr-verify.yml
saraswatpuneet: Ideally we would update the version of snapshot before releasing new one so that might take care of it for now

java/lib/build.gradle.kts
```diff
@@ -0,0 +1,77 @@
...
+dependencies {
+    // These dependencies are used only for junit tests
+    testImplementation("org.junit.jupiter:junit-jupiter:5.9.1")
+    testImplementation("io.github.hakky54:logcaptor:2.9.0")
+
+    // This dependency is exported to consumers, that is to say found on their compile classpath.
+    api("org.apache.commons:commons-math3:3.6.1")
+
+    // These dependencies are used internally, and not exposed to consumers on their own compile classpath.
+    implementation("com.google.guava:guava:31.1-jre")
```
aramikm, line 37: i don't think we need these. they were included in gradle generate template as an example

java/lib/build.gradle.kts
```diff
@@ -0,0 +1,77 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    // Apply the kotlin-jvm plugin for Kotlin JVM projects.
+    kotlin("jvm") version "1.6.21"
+    // Apply the maven-publish plugin for publishing the library.
+    id("maven-publish")
```
aramikm, line 7: we might need these two also. Signing one might be necessary if we are going to publish on maven central
id("java-library")
id("signing")

java/settings.gradle.kts
```diff
@@ -9,8 +9,8 @@
 
 plugins {
     // Apply the foojay-resolver plugin to allow automatic download of JDKs
-    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.4.0'
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.4.0"
 }
 
-rootProject.name = 'graphSdk'
-include('lib')
+rootProject.name = "graphSdk"
```
aramikm, line 15: let's rename this to `graph-sdk` for maven name convention compatibility

PR #93, opened 2023-06-01
`Kotlin` tests with `io.amplica.graphsdk.lib`

saraswatpuneet:
# Goal
The goal of this PR is to set a kotlin test project to test the published `io.amplica.graphsdk.lib`

# Checklist
- [x] Tests added


java-lib-test/src/test/kotlin/io/amplica/graphsdk/LibraryTest.kts
```diff
@@ -0,0 +1,43 @@
...
+
+    @Test
+    fun initiateRococoStateShouldWork() {
+        // act
+        val graph = Graph(Configuration.getRococo())
+
+        // assert
+        assertNotEquals(0, graph.unsafeNativeHandleWithoutGuard())
+        graph.finalize()
+    }
```
saraswatpuneet, line 42: Need to add more tests but for start we are able to use library

Makefile
```diff
@@ -66,6 +66,9 @@ test-ffi:
 test-jni: build-jni
 	@( cd java ; ./gradlew test --rerun-tasks)
 
+test-java-lib:
+	@( cd java-lib-test ; ./gradlew test --rerun-tasks)
```
aramikm, line 70: Can we move `java-lib-test` inside `java`? It doesn't have to be included inside upper gradle since it is just a small test client

java-lib-test/src/test/kotlin/io/amplica/graphsdk/LibraryTest.kts
aramikm: I don't think we need more tests. As long as the connection is established between `Current library -> Published Java Library -> Rust` it achieves it's purpose.

aramikm (commented):
After fixing the token issue now I'm seeing the following
` java.lang.UnsatisfiedLinkError`
not sure if I'm missing something.

java-lib-test/settings.gradle.kts
```diff
@@ -0,0 +1,10 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ *
+ * The settings file is used to specify which projects to include in your build.
+ *
+ * Detailed information about configuring a multi-project build in Gradle can be found
+ * in the user manual at https://docs.gradle.org/8.1.1/userguide/multi_project_builds.html
+ */
+
+rootProject.name = "java-lib-test"
```
aramikm, line 10: Should this be `example-client` or something like that?

saraswatpuneet: calling it `example-graphsdk-client`

java-lib-test/src/test/kotlin/io/amplica/graphsdk/LibraryTest.kts
saraswatpuneet: Lib is working as expect might not need to repeat all the tests

aramikm (approved):
Looks good! I'm using the same Gradle version 8.1.1. Could it be that you are running on linux and I'm testing it on Mac?

Issue #14, opened 2023-04-04
[GraphSdk] Jvm Bridge


aramikm:
## Why
We need a JVM based bridge to be able use graph sdk inside a jvm application (java/kotlin/android)

## Acceptence Criterea
- implement JNI function signatures for all graphsdk interface methods
- expose informative errors to JNI consumers
- Test graphSDK from inside a spring app (java/kotlin)
- Test graphSDK from inside an android app
- Test graphSDK from inside a scala app

# Subissues

- [x] JNI and Java project scaffolding #63 
- [x] Logging #64 
- [x] Errors #65 
- [x] API types #82
- [x] API definitions #83
- [x] Api Logic #84
- [x] Publish JNI #68

PR #94, opened 2023-06-01
some more safety related checks and mechanism

aramikm:
# Goal
The goal of this PR is to ensure that there are no race conditions between reading and writing state and Garbage collection during JNI api execution

related to #14


saraswatpuneet (approved):
lgtm!

PR #113, opened 2023-06-06
edge-case: handles reimports after updates

### LINKED ISSUE #49: [GraphSDK] [EdgeCase] Reimporting graph after having some updates inside UpdateTracker
## Steps to reproduce
1- Import the graph
2- Add new connections / Remove Connections
3- Reimport the graph again but this time the graph will already have those connections and basically have conflicts with Some of the updates inside **UpdateTracker**

## Acceptence Criterea
- On each graph import deconflict updates in tracker with graph
- Add tests

aramikm:
# Goal
The goal of this PR is to handle the edge case which re-imports of user data might happen after adding some updates. This would make sure that those updates are compatible with newly reimported user data and remove the non compatible ones.

Closes #49

# Notes
- Handles some of the unwraps around locks that were lingering around. We still have some unwraps in methods that do not return Result types but to handle them we will need to change the method signature which is  bigger than the scope of this PR.

# Checklist
- [x] Tests added

### DIFF
--- a/core/src/api/api.rs
+++ b/core/src/api/api.rs
@@ -55,8 +55,10 @@ use crate::{
 	},
 	frequency::Frequency,
 	graph::{
-		key_manager::UserKeyProvider,
-		shared_state_manager::{PriProvider, PublicKeyProvider, SharedStateManager},
+		key_manager::{UserKeyProvider, USER_KEY_MANAGER},
+		shared_state_manager::{
+			PriProvider, PublicKeyProvider, SharedStateManager, SHARED_STATE_MANAGER,
+		},
 		updates::UpdateEvent,
 		user::UserGraph,
 	},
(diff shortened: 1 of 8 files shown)


core/src/api/api.rs
```diff
@@ -269,7 +275,7 @@ impl GraphAPI for GraphState {
 		Ok(self
 			.shared_state_manager
 			.read()
-			.unwrap()
+			.map_err(|_| DsnpGraphError::FailedtoReadLock(SHARED_STATE_MANAGER.to_string()))?
```
saraswatpuneet, line 278: Do we want to also return error string or something here?

saraswatpuneet (approved):
lgtm! one comment

core/src/graph/updates.rs
```diff
@@ -95,6 +95,28 @@ impl UpdateTracker {
...
+	) {
+		if let Some(arr) = self.updates.get(&schema_id) {
+			let mut synced_updates = arr.clone();
+			synced_updates.retain(|e| match e {
+				UpdateEvent::Add { dsnp_user_id, .. }
+					if existing_connections.contains(&dsnp_user_id) =>
+					false,
+				UpdateEvent::Remove { dsnp_user_id, .. }
+					if !existing_connections.contains(&dsnp_user_id) =>
+					false,
```
JoeCap08055, line 111: 💯 Very cool... I wasn't aware of this type of match pattern

core/src/api/api.rs
aramikm: I'm not sure if I understand. Would you please explain more?

PR #122, opened 2023-06-12
[GraphSdk Node] Publish `@dsnp/graph-sdk` to npm and add some api code

### LINKED ISSUE #68: Package publish

### LINKED ISSUE #118: [graph-sdk] Publish npm package for JS/TS bindings

saraswatpuneet:
# Goal
The goal of this PR is to enable publishing of npm package while adding some simple api calls for reference

Note: The published package is not tested but that will come later (should work though as tested with jest)

Closes #118 

# Checklist
- [x] Tests added


saraswatpuneet:
image

.gitignore
```diff
@@ -17,5 +17,5 @@ bridge/ffi/src/c_example/libdsnp_graph_sdk_ffi.a
 bridge/ffi/src/c_example/main
 bridge/ffi/src/c_example/dsnp_graph_sdk_ffi.h
 *.class
-bridge/node/dsnp_graph_sdk.node
-bridge/node/js/dsnp_graph_sdk.node
+bridge/node/grapsdk.node
+bridge/node/js/grapsdk.node
```
JoeCap08055: Why the name change? Don't most of our other crates/packages for GraphSDK use the "dsnp-graph-*" pattern?

bridge/node/src/api.rs
```diff
@@ -33,33 +35,59 @@ pub fn get_graph_config(mut cx: FunctionContext) -> JsResult<JsObject> {
...
+	let rust_environment = unsafe { environment_from_js(&mut cx, environment_obj) };
+	let graph_state = GraphState::new(rust_environment);
+
+	// Generate a unique identifier for the graph state
+	let graph_state_id = {
+		let mut states = GRAPH_STATES.lock().unwrap();
+		let next_id = states.len();
+		states.insert(next_id, Arc::new(Mutex::new(graph_state)));
+		next_id
+	};
```
JoeCap08055, line 55: Why are we managing this differently than FFI and JNI? There we just return the pointer (`jlong` for JNI); here we're returning an index into the Vec... which is not guaranteed to be constant, if graph states are removed...

(Personally this is why I advocated for managing graph states using a HashMap instead; since the pointer is opaque to the caller anyway, a hash key would be more manageable, and wouldn't require all the work we currently do with box/unbox, etc)

JoeCap08055: The insertion logic now doesn't work with a hashmap. You can't use `states.len()` as the next ID; you'll wind up with hash key clashes. I'd say use a monotonically-increasing value (with maybe adding a wraparound later).

bridge/node/src/api.rs
```diff
@@ -33,33 +43,69 @@ pub fn get_graph_config(mut cx: FunctionContext) -> JsResult<JsObject> {
...
+pub fn initialize_graph_state(mut cx: FunctionContext) -> JsResult<JsNumber> {
+	let environment_obj = cx.argument::<JsObject>(0)?;
+	let rust_environment = unsafe { environment_from_js(&mut cx, environment_obj) };
+	let graph_state = GraphState::new(rust_environment);
+
+	// Generate a unique identifier for the graph state
+	let graph_state_id = {
+		let mut next_id = NEXT_GRAPH_STATE_ID.lock().unwrap();
+		let id = *next_id;
+		*next_id += 1;
```
JoeCap08055, line 61: Maybe use `.wrapping_add(1)` here? Reconnection service will ideally be long-running, and as we load/unload user  graphs we'll burn through ids at a rate of potentially `(num_users)**2`. (Realizing it's unlikely we'll overflow a `usize` before that app is restarted at some point, but it's probably even less likely that we'll have a key conflict if we wrap (since our intended usage pattern for graph is to load a graph, do stuff with it, then remove it from the cache at some point)

saraswatpuneet: Thats a great suggestion

JoeCap08055 (approved):

PR #123, opened 2023-06-14
Define TS types and Native interface for graphsdk ts

### LINKED ISSUE #121: [graph-sdk] Add Typescript types/package for Neon NodeJS bindings

saraswatpuneet:
# Goal
The goal of this PR is to define TS types and Native interface for graphsdk ts

Closes #121 

# Checklist
- [x] Tests added


bridge/node/js/graph.test.ts
```diff
@@ -1,20 +1,37 @@
...
+    expect(config.sdkMaxUsersGraphSize).toEqual(1000);
+});
+
+test('getGraphConfig with Rococo environment should return the graph config', async () => {
+  const environment: EnvironmentInterface = { environmentType: EnvironmentType.Rococo };
+  const graph = new Graph(environment);
+  const config = await graph.getGraphConfig(environment);
+  expect(config).toBeDefined();
+  expect(config.sdkMaxUsersGraphSize).toEqual(1000);
+});
```
saraswatpuneet, line 37: Some tests around APIs that are currently working

bridge/node/js/graph.ts
```diff
@@ -1,25 +1,111 @@
...
+            const registry = new FinalizationRegistry<Graph>(() => {
+                finalizer();
+            });
+            registry.register({}, this);
+        } else if (typeof process !== "undefined" && typeof process.on === "function") {
+            process.on("exit", finalizer);
+        } else {
+            console.warn("Unable to register finalizer. Memory may not be freed correctly.");
+        }
+    }
```
saraswatpuneet, line 102: Finalizer pattern to free underlying graphstate upon garbage collection

bridge/node/README.md
```diff
@@ -0,0 +1,90 @@
+# DSNP Graph SDK
```
JoeCap08055, line 1: Nice doc. Might just want to call out that interacting with the chain (fetching graph to import, committing exported graph changes to the chain) would be done using another package such as `@polkadot-js`

bridge/node/js/graph.ts
```diff
@@ -1,25 +1,110 @@
...
+
+    public async freeGraphState(): Promise<void> {
```
JoeCap08055, line 103: I wonder if, since we're using a registered finalizer, we should make this function private--otherwise you could call `freeGraphState` and then this object would be left with a dangling handle.

saraswatpuneet: Ideally it wont happen given state will be remove from hashmap and calling function will just get an error, but will hide it

bridge/node/README.md
saraswatpuneet: Yes intend to do it with #124 PR which will have full example/tests etc.

bridge/node/js/graph.ts
saraswatpuneet: Apparently finalizer wont free the state tried few examples, for now we will explicitly free the state and mention that in documentation

JoeCap08055: I'm not sure I understand... so are you saying the registered finalizer doesn't work? If so, should we just remove that?

saraswatpuneet: Yes it does not work, i think it would need to know the actual State object to be able to track it and call garbage collection on it, but we hide the state and manage it on rust side, so we have to explicitly garbage collect hidden state

saraswatpuneet: I removed the finalizer for now, we would call free state explicitly as a start?

JoeCap08055: That seems fine for now. I'll put a story in the backlog to investigate getting the finalizer to work.

PR #131, opened 2023-06-21
[WIP] Api to Generate x25519 keys

### LINKED ISSUE #127: [GraphSDK] create a new API endpoint to be able to generate X25519 keys
# Why
When doing integrations with Saas client it became obvious that they only have access to Sr25519 keys and can not generate any X25519 keys which we use for the graph. All alternative JVM libraries that generate x25519 are heavy weight and not that great so it makes sense to provide this functionality as an api in GraphSdk to allow generating new x25519 keys without adding a new library.

# What
- [x] Add a new api endpoint `generate_keypair(key_type: GraphKeyType)` to graph sdk which does not require state. Please follow the same pattern as **deserialize_dsnp_keys** endpoint which does not have `&self` as a param.
- [x] Add JNI, FFI, Node bridges
- [x] Add integration test in Rust/Java/FFI/Node

# Notes
- It seems like that wallet currently does not support x25519 and we have to communicate with wallet teams to add the support for those type of keys
- Ensure that keys being generated are unique and use randomness provided by OS or device. One option to just use `StackKeyPair::gen()` from **dryoc** library which is using the same implementation as lib-sodium

saraswatpuneet:
# Goal
The goal of this PR is  to provider an api to generate `GraphKeyType` keypairs and return 
Closes #127 

# Checklist
- [ ] Tests added

### DIFF
--- a/bridge/node/js/graph.test.ts
+++ b/bridge/node/js/graph.test.ts
@@ -299,7 +299,7 @@ test('deserializeDsnpKeys with empty keys should return empty array', async () =
         keysHash: 100,
         keys: [],
     } as DsnpKeys;
-    const connections = await graph.deserializeDsnpKeys(keys);
+    const connections = await Graph.deserializeDsnpKeys(keys);
     expect(connections).toBeDefined();
     expect(connections.length).toEqual(0);
     await graph.freeGraphState();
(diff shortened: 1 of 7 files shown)


saraswatpuneet:
Do we need it for ffi? libsodium is very well supported in those

bridge/node/js/graph.test.ts
```diff
@@ -299,7 +299,7 @@ test('deserializeDsnpKeys with empty keys should return empty array', async () =
         keysHash: 100,
         keys: [],
     } as DsnpKeys;
-    const connections = await graph.deserializeDsnpKeys(keys);
+    const connections = await Graph.deserializeDsnpKeys(keys);
```
JoeCap08055, line 302: nit: this is confusing to call it `connections` instead of `keys`

JoeCap08055:
> Do we need it for ffi? libsodium is very well supported in those

I think it would be useful to have it so that a client app can generate keys without worrying about the correct parameters, etc. But we can skip it for now; it'd be easy enough to add later if we want.

JoeCap08055 (approved):

PR #135, opened 2023-06-27
Increase graph limit to 10k

### LINKED ISSUE #134: uodate connetion limit on configs
update to 10k

JoeCap08055:
# Goal
The goal of this PR is to increase the number of user graphs that can be held in a single `GraphState` in order to support holding all of at least one user's connections in one `GraphState`.

Closes #134 

# Checklist
- [x] Tests updated


saraswatpuneet (approved):
Lgtm! We can circle back with aramik soon

PR #141, opened 2023-07-03
add schemaid field for public key in `Config`

### LINKED ISSUE #139: [graph-sdk] Add Itemized SchemaId for Graph public keys to GraphSDK Config structure
In order to interact with Private Graph, we need to be able to retrieve a user's public graph key from the chain. The SchemaId for this Itemized schema needs to be stored in the Config for the environment (Rococo, Mainnet).

Acceptance critieria:
* Graph key schema id preset in Config for mainnet, rococo
* Graph key schema id retrievable in all APIs and bridges (FFI, JNI, Node)

saraswatpuneet:
# Goal
The goal of this PR is to add a field in config to hold the value of schemaid for itemized public key storage on frequency

Closes #139


bridge/jni/src/mappings.rs
```diff
@@ -312,6 +314,7 @@ fn map_config_to_rust(config: proto_output::Config) -> SdkJniResult<RustConfig>
 		max_key_page_size_bytes: config.max_key_page_size_bytes,
 		dsnp_versions,
 		schema_map,
+		graph_public_key_schema_id: config.graph_public_key_schema_id as SchemaId,
```
JoeCap08055, line 317: nit: don't think you need the `as SchemaId` here; `Config.graph_public_key_schema_id` _is_ a `SchemaId`

saraswatpuneet: this is probably confusing but here  `config` is protobuf config where graph key schema id is of type `uint32`

JoeCap08055: Ah, okay

aramikm, line 317:
```suggestion
		graph_public_key_schema_id: SchemaId::try_from(config.graph_public_key_schema_id).map_err(|_| SdkJniError::InvalidRequest("invalid SchemaId"))?,
```

bridge/common/src/proto_types/output.rs
aramikm, line 1: Why unrelated Rust and Java generated files are changed? is it because of `protoc` version?

aramikm (approved):
Looks good just one suggestion and one question.

bridge/common/src/proto_types/output.rs
saraswatpuneet: Yes its probably that and also a new field is added to config so had to re-generate these files

bridge/jni/src/mappings.rs
```diff
@@ -194,6 +194,8 @@ pub fn serialize_config<'local>(
 		max_graph_page_size_bytes: config.max_graph_page_size_bytes,
 		dsnp_versions: map_dsnp_versions_to_proto(&config.dsnp_versions)?,
 		schema_map: map_schema_map_to_proto(&config.schema_map)?,
+		graph_public_key_schema_id: u32::try_from(config.graph_public_key_schema_id)
```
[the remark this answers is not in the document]
JoeCap08055: Right, but we can choose to validate more narrowly than the incoming protobuf, right? `u16` is a subset of `u32`. Or do you think we should allow it and fail higher in the stack?

PR #143, opened 2023-07-06
[DARFT] use build.rs to generate proto files

### LINKED ISSUE #107: [GraphSDK] generate proto classes using build.rs
# Story
Currently we are included generated proto class in git and we need to make it part of them build process

# Accpetence criterea
- Use build.rs in jni to build rust and java proto class
- remove generated classes from git

aramikm:
# Goal
The goal of this PR is to use `build.rs` script to generate rust and java files for proto. build.rs file is calling `make build-protos` command anytime one of proto files changed

Closes #107 

# Discussion
- removes generated files from repo
- uses `build.rs` for Rust and Java
- I'm not sure if the effort to remove the generated Java and Rust files from git does not justify the benefits. Specially since we saw some recent changes in protoc versions and their Rust support.
- Had to change the previous approach which was using `protobuf-codegen` due to that library having some bugs which was not allowing to use additional flags which was necessary to allow optional experimental fields


bridge/common/build.rs
```diff
@@ -0,0 +1,34 @@
...
+	// remove old generated files
+	let paths = ["src/proto_types", "../../java/lib/src/main/java/io/amplica/graphsdk/models"];
+	for path in paths {
+		let _ = fs::remove_dir_all(path);
+		let _ = fs::create_dir(path);
+	}
+
+	// generate new Rust files
+	protobuf_codegen::Codegen::new()
+		.protoc()
```
aramikm, line 17: should detect the correct installed protoc based on platform, so we don't need to care about the platform

bridge/common/build.rs
```diff
@@ -0,0 +1,34 @@
...
+		.includes(&["protos"])
+		.input("protos/input.proto")
+		.input("protos/output.proto")
+		.out_dir("src/proto_types")
+		.run_from_script();
+
+	// generate new Java files
+	Command::new("protoc")
```
aramikm, line 25: protoc for java worked regardless of the platform so we don't need to care about the platform specific path.

java/lib/src/main/java/io/amplica/graphsdk/models/Actions.java
```diff
@@ -217,7 +217,7 @@ protected java.lang.Object newInstance(
```
JoeCap08055, line 217: Shouldn't all of these generated files be deleted from the repo now that they're being generated in the build process?

bridge/common/build.rs
```diff
@@ -0,0 +1,20 @@
...
+	println!("cargo:rerun-if-changed=build.rs");
+	println!("cargo:rerun-if-changed=protos/input.proto");
+	println!("cargo:rerun-if-changed=protos/output.proto");
+
+	// generate proto files
+	let output = Command::new("make")
+		.current_dir("../..")
+		.args(["build-protos"])
+		.output()
+		.expect("failed to execute build-proto");
```
JoeCap08055, line 13: You mentioned in the PR description that we're changing the approach from relying on `protobuf-codegen`, but the Makefile hasn't changed and here we're just invoking the Makefile. If that's the case, we could just update the Makefile with the proper dependencies; `build.rs` isn't really buying us anything.

Aren't there a couple Rust crates that embed protoc generation without proxying to CLI?

Also, I'm not generally a fan of including system tool setup in the Makefile (ie, `make install-protos`). You don't install the Rust, Java or C compilers that way. I think the installation of the protobuf packages should be part of the dev environment or image setup. (CI Runner image setup for pipeline; README instructions for local dev environment setup)

JoeCap08055: Ah, looks like a previous commit used the crate-based method for generating protos, but you got rid of it. Why?

aramikm: I mentioned in the PR discussions section. But the main reason was some bugs in the library that was not allowing to use additional flags which was necessary to allow optional experimental fields.

The benefit of `build.rs` is that if any of the protos changed when you rebuild the project it will regenerate the necessary changed files and you don't have to run the build-protos manually after changing any of them. This will automate any changes and reduces the chance of forgetting.

As for removing the generated files form git I'm on the edge since I've encountered some issues and not convinced if the benefits would worth the time. But I don't have any strong opinions on this and open to give it another shot since we are doing most of the things that allows it to happen anyway.

aramikm: As for using the pure rust protoc implementations, as far as I've checked they only support generating rust code and since we need to generate java also it does make sense to only use the official protoc which can generate in most languages.

JoeCap08055: I'm not a fan of circular tool invocation (make -> cargo -> build.rs -> make). Using `build.rs` doesn't buy us anything, IMO, and generally if we're using a Makefile-based build system on top of cargo, let's use it:

```
PROTO_DIR := bridge/common/protos
RUST_PROTO_DIR := bridge/common/src/proto_types
JAVA_PROTO_DIR := java/lib/src/main/java/io/amplica/graphsdk/models
PROTOS := $(wildcard $(PROTO_DIR)/*.proto)
RUST_PROTOS := $(foreach proto,$(PROTOS), $(patsubst %.proto,$(RUST_PROTO_DIR)/%.rs,$(notdir $(proto)))) $(RUST_PROTO_DIR)/mod.rs

.PHONY: build-protos
build-protos: build-rust-protos $(JAVA_PROTO_DIR)

.PHONY: build-rust-protos
build-rust-protos: $(RUST_PROTOS)

$(RUST_PROTO_DIR)/%.rs: $(PROTO_DIR)/%.proto
	@echo "Generating Rust protobuf types..."
	@$(PROTOC) --rust_out $(RUST_PROTO_DIR) $(PROTOS)

$(JAVA_PROTO_DIR): $(PROTOS)
	@echo "Generating Java protobuf types..."
	@rm -f ./java/lib/src/main/java/io/amplica/graphsdk/models/*
	@$(PROTOC) --java_out ./java/lib/src/main/java/ ./bridge/common/protos/input.proto ./bridge/common/protos/output.proto
```

(Note, the above is not _quite_ right for the Java side of things... I think we might want to consider changing the option `option java_multiple_files = false` to simplify dependencies.)

aramikm:
Closing a stale PR

PR #146, opened 2023-07-12
Make dsnp keys optional

JoeCap08055:
# Goal
The goal of this PR is to make DSNP keys optional in graph import bundle payloads when importing to a graph.

Closes 

# Checklist
- [ ] Tests added


core/src/util/builders.rs
```diff
@@ -235,7 +235,10 @@ impl ImportBundleBuilder {
 		};
 
 		ImportBundle {
-			dsnp_keys: DsnpKeys { keys, keys_hash, dsnp_user_id: self.dsnp_user_id },
+			dsnp_keys: match keys.len() {
+				0 => None,
+				_ => Some(DsnpKeys { keys, keys_hash, dsnp_user_id: self.dsnp_user_id }),
+			},
```
saraswatpuneet, line 241: This probably needs to be done in `ImportBundleBuilder.ts`

saraswatpuneet (approved):
lgtm!  I think import bundle builder will need updates otherwise all good :+1:

core/src/util/builders.rs
JoeCap08055: In `import-bundle-builder.ts`, it simply allows `dsnpKeys` to be undefined (previously was inserting a dummy default object if undefined)

saraswatpuneet (approved):

PR #148, opened 2023-07-13
dry run PR publish

### LINKED ISSUE #147: [graph-sdk] CI pipeline should not publish packages from PR branches
The CI pipeline should only publish packages (npm, maven) on a commit or merge to `main`.

Problem: Currently, every commit to a PR causes a package publishing attempt with the same version number (for npm this fails due to npm repo policy).

Acceptance criteria: Packages are only published on merge/commit to `main`

saraswatpuneet:
# Goal
The goal of this PR is to not publish npm graphsdk package for PRs, to verify on PR a `--dry-run` is performed on package to be published 

Closes #147

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches:
       - main
+  push :
+    branches:
+      - main
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -290,7 +290,14 @@ jobs:
         run: npm run native:build-release && cp README.md dist/README.md
       - name: Publish NPM Package
         working-directory: bridge/node
+        if: github.ref == 'refs/heads/main'
         run: npm publish --access public
+        env:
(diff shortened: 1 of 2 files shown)


saraswatpuneet:
@JoeCap08055  it just publishes on main build upon merge, so any branch !main wont publish but upon merge it will for Java and Jni we are just publishing to github for now, so it should be ok, but we have story for maven central etc. that should take care of that

.github/workflows/pr-verify.yml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches:
       - main
+  push :
+    branches:
+      - main
```
JoeCap08055, line 11: re: my previous comment, it seems like this is limiting the entire pipeline to:
* PRs that target `main` as the destination branch
* Pushes directly to `main` or merges to `main`

So if I have a PR on a sub-branch, the pipeline won't run at all:
branches: `main/branch1` and `main/branch_1/branch_1a`
PR on `main/branch_1a` -> `main/branch_1` won't run CI in the PR itself because the target branch is not `main`
Merging the PR will not run CI because the branch being merged to is not `main`

Not a super big deal as we don't tend to do sub-branches, but food for thought.

saraswatpuneet: Yeah i mean its a simple pipelines that does the job for now

PR #152, opened 2023-07-14
Enhance builder to take multiple keypairs

saraswatpuneet:
For private graph there will be a list of keys coming from provider in general

### DIFF
--- a/bridge/node/js/import-bundle-builder.spec.ts
+++ b/bridge/node/js/import-bundle-builder.spec.ts
@@ -29,7 +29,7 @@ describe('ImportBundleBuilder', () => {
     const importBundleBuilder = new ImportBundleBuilder()
       .withDsnpUserId(dsnpUserId)
       .withSchemaId(schemaId)
-      .withGraphKeyPair(GraphKeyType.X25519, new Uint8Array([1, 2, 3]), new Uint8Array([4, 5, 6]))
+      .withGraphKeyPair([{keyType: GraphKeyType.X25519, publicKey: new Uint8Array([1, 2, 3]), secretKey: new Uint8Array([4, 5, 6])} as GraphKeyPair])
       .withDsnpKeys(dsnpKeys)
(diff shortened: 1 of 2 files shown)


bridge/node/js/import-bundle-builder.ts
```diff
@@ -25,9 +25,9 @@ export class ImportBundleBuilder {
     return new ImportBundleBuilder({ ...this.values, schemaId });
   }
 
-  public withGraphKeyPair(keyType: GraphKeyType, publicKey: Uint8Array, secretKey: Uint8Array): ImportBundleBuilder {
```
JoeCap08055, line 28: nit: would be nice to have 2 variants: `withGraphKeyPair(key: GraphKeyPair)` and `withGraphKeyPairs(keys: GraphKeyPair[])`

saraswatpuneet: yeah thats cool i like it, and wont break any consumer if any

bridge/node/js/import-bundle-builder.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('ImportBundleBuilder', () => {
     const importBundleBuilder = new ImportBundleBuilder()
       .withDsnpUserId(dsnpUserId)
       .withSchemaId(schemaId)
-      .withGraphKeyPair([{keyType: GraphKeyType.X25519, publicKey: new Uint8Array([1, 2, 3]), secretKey: new Uint8Array([4, 5, 6])} as GraphKeyPair])
+      .withGraphKeyPairs([{keyType: GraphKeyType.X25519, publicKey: new Uint8Array([1, 2, 3]), secretKey: new Uint8Array([4, 5, 6])} as GraphKeyPair])
```
JoeCap08055, line 32: nit: tests should cover both variants of the function (single object & array)

JoeCap08055 (approved):

PR #157, opened 2023-07-18
fix: update mainnet config with actual published schema ids

### LINKED ISSUE #104: [GraphSDK] check in SDK and publish Mainnet Schemas
# Story
Publish graph related schemas on main-net and update Graph SDK with registered schema Ids.
Check and finalize mainnet sdk settings

JoeCap08055:
# Goal
* Update mainnet config with published schema ids
* Remove public friendship from config
 
Closes #104 

# Checklist
- [ ] Tests added

### DIFF
--- a/config/resources/configs/frequency.json
+++ b/config/resources/configs/frequency.json
@@ -4,10 +4,10 @@
   "maxGraphPageSizeBytes": 1024,
   "maxPageId": 16,
   "maxKeyPageSizeBytes": 65536,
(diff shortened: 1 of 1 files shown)


saraswatpuneet (approved):

wilwade (approved):

aramikm (approved):
looks good, the test needs to be fixed and what would happen if somebody tries to use public friendship after removing that config? Maybe we need to check against it in our validators

PR #161, opened 2023-07-27
update api to handle empty pages

saraswatpuneet:
Let neon accept empty pages

### DIFF
--- a/bridge/node/src/helper.rs
+++ b/bridge/node/src/helper.rs
@@ -299,8 +299,11 @@ pub fn import_bundle_from_js_object<'a, C: Context<'a>>(
 		None => Vec::new(),
 	};
 
-	let pages: Handle<'_, JsArray> = import_bundle_js.get(cx, "pages")?;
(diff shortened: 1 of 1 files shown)


bridge/node/src/helper.rs
```diff
@@ -299,8 +299,11 @@ pub fn import_bundle_from_js_object<'a, C: Context<'a>>(
 		None => Vec::new(),
 	};
 
-	let pages: Handle<'_, JsArray> = import_bundle_js.get(cx, "pages")?;
-	let pages: Vec<PageData> = pages_from_js(cx, pages)?;
+	let pages_res: Handle<'_, JsArray> = import_bundle_js.get(cx, "pages")?;
+	let mut pages: Vec<PageData> = Vec::new();
+	if pages_res.len(cx) > 0 {
+		pages = pages_from_js(cx, pages_res)?;
+	}
```
saraswatpuneet, line 306: Non breaking change just when we have empty pages we would not go that route

PR #167, opened 2023-07-29
Incremented snapshot jvm version

aramikm:
# Goal
The goal of this PR is to use a new incremented snapshot for jvm


saraswatpuneet (approved):

JoeCap08055:
At what point do we stop with the `-SNAPSHOT` and cut a release?

aramikm:
@JoeCap08055 We should figure it out but this one was to allow custodial wallet changes merge and having any dependency on a snapshot version

PR #170, opened 2023-08-03
ImportBundles: set condition to allow importing `dsnpKeys` and `keyPairs` without schemaId

### LINKED ISSUE #168: [bug] Unable to import DSNP public keys without schema ID, and import with empty pages still clears graph state
1. As a client of the GraphSDK API, I should be able to import a user's public DSNP graph keys ONLY, without importing the entire graph. Currently the import function assumes that `schema_id` is present in the import bundle

2. If I import only public DSNP  keys with a zero-length vector of pages, the existing graph state for the user is still cleared. I should be able to import only public keys; If no pages are present in the import bundle, the graph state should be untouched. _(Note: if it is desired to be able to clear a user graph, it could be removed and re-added to the graph state, or an additional `clear_user_graph` method could be created)_

saraswatpuneet:
# Goal
The goal of this PR is to set condition to allow importing `dsnpKeys` and `keyPairs` without schemaId

Closes #168 

# Checklist
- [x] Tests added

### DIFF
--- a/bridge/node/js/graph.test.ts
+++ b/bridge/node/js/graph.test.ts
@@ -70,423 +70,438 @@ describe("Graph tests", () => {
     graph.freeGraphState();
   });
 
-  test('printHelloGraph should print "Hello, Graph!"', async () => {
-    // Mock the console.log function
-    const consoleLogMock = jest.spyOn(console, "log").mockImplementation();
-    graph.printHelloGraph();
-    expect(consoleLogMock).toHaveBeenCalledWith("Hello, Graph!");
-  });
+  // test('printHelloGraph should print "Hello, Graph!"', async () => {
+  //   // Mock the console.log function
+  //   const consoleLogMock = jest.spyOn(console, "log").mockImplementation();
+  //   graph.printHelloGraph();
+  //   expect(consoleLogMock).toHaveBeenCalledWith("Hello, Graph!");
+  // });
 
-  test("getGraphConfig should return the graph config", async () => {
-    const config_ret = graph.getGraphConfig(environment);
-    expect(config_ret).toBeDefined();
-    expect(config_ret.graphPublicKeySchemaId).toEqual(11);
-  });
+  // test("getGraphConfig should return the graph config", async () => {
+  //   const config_ret = graph.getGraphConfig(environment);
+  //   expect(config_ret).toBeDefined();
+  //   expect(config_ret.graphPublicKeySchemaId).toEqual(11);
+  // });
 
-  test("getGraphConfig with Mainnet environment should return the graph config", async () => {
-    const environment: EnvironmentInterface = {
-      environmentType: EnvironmentType.Mainnet,
-    };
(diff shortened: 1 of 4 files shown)


core/src/api/api.rs
```diff
@@ -417,11 +417,20 @@ impl GraphState {
...
+				Some(connection_type) => connection_type,
+				None => {
+					if pages.is_empty() && !key_pairs.is_empty() && !dsnp_keys.is_none() {
+						// This condition implies that only keys are being imported
+						ConnectionType::Follow(PrivacyType::Private)
+					} else {
+						return Err(DsnpGraphError::InvalidSchemaId(*schema_id))
+					}
+				},
+			};
```
saraswatpuneet, line 432: Here the very specific condition is checked
if we dont have schemaId and pages are empty then if and only if `dsnpKeys` and `keyPairs` are provided, we proceed else throw the error

saraswatpuneet: In this part, its more like a safe bypass for only importing the keys

bridge/node/js/graph.test.ts
```diff
@@ -483,4 +488,62 @@ describe("Graph tests", () => {
...
+      ownerDsnpUserId: "1000",
+      dsnpKeys,
+      connection: connectionPrivate,
+    }
+    const appliedAction = await  graph.applyActions([privateConnectAction]);
+    expect(appliedAction).toEqual(true);
+    const exported = graph.exportUpdates();
+    expect(exported).toBeDefined();
+    expect(exported.length).toEqual(1);
+  });
```
saraswatpuneet, line 548: This test, import bundles with empty pages, 0 schemaId and then create a private connection to ensure we are able to use those keys and export updates

core/src/api/api.rs
```diff
@@ -417,11 +417,20 @@ impl GraphState {
...
-				.get_connection_type_from_schema_id(*schema_id)
-				.ok_or(DsnpGraphError::InvalidSchemaId(*schema_id))?;
+			let connection_type_option =
+				self.environment.get_config().get_connection_type_from_schema_id(*schema_id);
+			let connection_type = match connection_type_option {
+				Some(connection_type) => connection_type,
+				None => {
+					if pages.is_empty() && (!key_pairs.is_empty() || !dsnp_keys.is_none()) {
+						// This condition implies that only keys are being imported
+						ConnectionType::Follow(PrivacyType::Private)
```
aramikm, line 427: I'm not sure about this, it doesn't look that clean

saraswatpuneet: I am divided here too, however this connection_type is not used anywhere in the rest of the logic under these conditions

We can skip it away if pages are empty? Looking into some cleaner solution

saraswatpuneet: Tweaked it a bit to be cleaner, let me know your thoughts

aramikm (approved):
Looks better. Thanks.

Issue #155, opened 2023-07-16
[graph-sdk] SPIKE: publish JAR with multi-architecture native component


JoeCap08055:
Currently, the published JAR file only supports the `linux/amd64` or `linux/x86_64` architecture, because that's the architecture of the GitHub CI runner that runs the package publish pipeline. In order to properly support JVM-based language developers, we need to support multiple architectures.

* Research existing projects that deploy JARs with native components. How do they handle multi-architecture?
* This would likely necessitate some modification of the `cargo` manifests to be able to cross-compile multiple different target architectures in a single build environment.
* Research WASM support in Java; is it possible to enable the `wasm` target for `graph-sdk` so that only a wasm binary is required. This may be necessary anyway to support browser & mobile...
* Other?

Desired output:
* Analysis & decision of an appropriate solution, with a working sample

aramikm:
This is done in #180

Issue #154, opened 2023-07-16
[graph-sdk] SPIKE: Publish multi-architecture prebuilt binaries for Node.js


JoeCap08055:
Currently, the published npm package `@dsnp/graph-sdk` only supports the `linux/amd64` or `linux/x86_64` architecture, because that's the architecture of the GitHub CI runner that runs the package publish pipeline. In order to properly support Node.js developers, we need to support multiple architectures. This could be achieved in several ways:

* Using something akin to [node-gyp](https://github.com/nodejs/node-gyp), which installs native modules from C/C++ source with installation-time compilation. Could this be adapted (or is there another tool) for Rust?
* Using [node-pre-gyp](https://github.com/mapbox/node-pre-gyp) to manage publishing pre-built binaries for some common architectures.
  * This would likely necessitate some modification of the `cargo` manifests to be able to cross-compile multiple different target architectures in a single build environment.
* Enable the `wasm` target for `graph-sdk` so that only a wasm environment is required. This may be necessary anyway to support browser & mobile...
* Other?

Desired output:
* Analysis & decision of an appropriate solution, with a working sample

## Notes:
See [this issue](https://github.com/neon-bindings/neon/issues/117) (especially the discussion).

saraswatpuneet:
I would say `wasm` would be ideal but due to unknown amount of work that it might need, `node-gyp` is used everywhere else e.g. `libsignal`

aramikm:
This is done in #180

PR #172, opened 2023-08-10
Github action for release

### LINKED ISSUE #102: [GraphSDK] Non-linux package support for JVM
# Story
Currently published packages only are linux compatible since it compiles the rust to dynamic library with `.so` extension and this limits the published package to only Linux/Unix subsystems.

# Acceptence criterea
- Find a way to publish non-linux graph sdk so we can use graph-sdk library inside a  JVM based langauge outside linux
   - this can be done either by generating .`dylib` for mac and `.dll` for windows inside a separate jar file and publish those jar along with the linux one
   - Or the dylib or dll can be uploaded separately and gradle can detect the environment and try to download them if not present
 
# Notes
- Can take a look at https://github.com/emeraldpay/polkaj
- https://github.com/AmplicaLabs/polkaj/blob/latest_release/polkaj-schnorrkel/src/main/java/io/emeraldpay/polkaj/schnorrkel/SchnorrkelNative.java#L116-L148
- https://github.com/signalapp/libsignal/blob/main/java/build.gradle#L54 for some patterns.
- might also be able to use Maven qualifiers

aramikm:
# Goal
The goal of this PR is to facilitate releases for graph-sdk

Closes #102
Related to  #155 
Related to #154

# Changes
- publishing pre-built binaries as release assets which gradle and node bridges can download using `make download-jni` and `make build-node-download` commands. The idea is that we currently only need non linux targets for developing.
- Used universal binaries for macOS native binaries supporting `aarch64-apple-darwin` and `x86_64-apple-darwin`

# Discussions
- Tried using that https://github.com/dherman/sniff-bytes/blob/main/.github/workflows/publish.yml for publishing multi-target packages but the inner actions used are all blocked and we would need to whitelist all of them .

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -76,26 +76,6 @@ jobs:
       - name: Cargo fmt
         run: make format-check
 
-  #  clippy:
-  #    name: Run clippy
-  #    runs-on: ubuntu-latest
-  #    steps:
-  #      - name: Checkout code
-  #        uses: actions/checkout@v3
-  #        with:
-  #          fetch-depth: 0
-  #
-  #      - name: Install Rust Toolchain
-  #        uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af
-  #        with:
-  #          toolchain: stable
-  #          default: true
-  #          profile: minimal
-  #          target: wasm32-unknown-unknown
-  #
-  #      - name: Cargo clippy
-  #        run: make clippy
-
   deny:
     name: Cargo Deny
     runs-on: ubuntu-latest
@@ -203,16 +183,9 @@ jobs:
           while read line; do
               echo "::warning title=Invalid file permissions automatically fixed::$line"
           done
-      - name: Upload Docs
-        uses: actions/upload-pages-artifact@v1
-        with:
-          path: ./target/doc
-      - name: Deploy Docs
-        uses: actions/deploy-pages@v2
-        id: deployment
 
   build_gradlew_jvm:
-    name: Build and Publish JVM
+    name: Build JVM
     runs-on: ubuntu-latest
     needs: [verify, test, lint, build]
     permissions:
@@ -230,12 +203,6 @@ jobs:
       - name: Build and Publish with Gradle
         working-directory: java
         run: ./gradlew build
-      - name: Publish to GitHub Packages
-        working-directory: java
-        run: ./gradlew publish
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GITHUB_ACTOR: ${{ github.actor }}
 
   build_test_node_bridge:
     name: Build and Test Node Bridge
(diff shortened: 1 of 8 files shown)


.github/workflows/pr-verify.yml
```diff
@@ -203,18 +203,11 @@ jobs:
           while read line; do
               echo "::warning title=Invalid file permissions automatically fixed::$line"
           done
-      - name: Upload Docs
```
aramikm, line 206: Only uploading on releases

.github/workflows/pr-verify.yml
```diff
@@ -274,65 +267,10 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_ACTOR: ${{ github.actor }}
 
-  publish_npm_package:
```
aramikm, line 277: Only publishing on release. Java behavior is a bit different since we are using snapshot version on that and it might be fine releasing snapshot versions to GitHub but that's also questionable. Open to discuss which way makes  more sense.

.github/workflows/pr-verify.yml
```diff
@@ -76,26 +76,6 @@ jobs:
       - name: Cargo fmt
         run: make format-check
 
-  #  clippy:
```
aramikm, line 79: enabled it but it failed and could not reproduce the error locally. It seems like it might be strictly dependent on rust version being used which might be different from my local and CI.

aramikm (commented):
It seems like that most of the code is duplicate between those two and maybe it makes sense to have only one release instead of gating it by bridge.

saraswatpuneet (changes requested):
looks good!,  lets have one more sync up

JoeCap08055 (approved):
looks good!

.github/workflows/release_java.yml
```diff
@@ -0,0 +1,216 @@
+name: Release Java
+run-name: Release ${{github.event.inputs.release-version || github.ref_name}}
+concurrency:
+  group: ${{github.workflow}}-${{github.ref}}
+  cancel-in-progress: true
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+-java" # ex. v1.0.0-java
+      - "v0.0.1-rc[0-9]+-java" # used for testing only
```
wilwade, line 10: I think I found the issue. The `\+` fails inside of the `"` due to yaml. Putting it into single quotes `'` works

```suggestion
      - 'v[0-9]+.[0-9]+.[0-9]+\+java[0-9]+' # ex. v1.0.0+java1
      - 'v0.0.1-rc[0-9]+\+java[0-9]+' # used for testing only

```

aramikm: Great discovery! I'm still surprised why that is the case.

.github/workflows/pr-verify.yml
aramikm: Going to use the same process as Node and only publish on release. No more snapshots!

PR #174, opened 2023-08-11
bump versions to 1.0.0

### LINKED ISSUE #173: [graph-sdk] bump version to v1.0.0
# Definition

We would like to have our first release with version v1.0.0 and this would make sure that all the dependencies have the same version.

aramikm:
# Goal
The goal of this PR is bump versions to v1.0.0 so we can release our first stable version.

Closes #173

### DIFF
--- a/java/lib/build.gradle.kts
+++ b/java/lib/build.gradle.kts
@@ -12,12 +12,12 @@ plugins {
 }
 
 group = "io.amplica.graphsdk"
-val uploadedBinariesVersion = "0.0.1-rc1"
+val uploadedBinariesVersion = "1.0.0"
 java.sourceCompatibility = JavaVersion.VERSION_17
 version = if (project.hasProperty("projVersion")) {
     project.properties["projVersion"]!!
 } else {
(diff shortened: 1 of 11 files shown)


java/lib/build.gradle.kts
```diff
@@ -12,12 +12,12 @@ plugins {
...
 
 group = "io.amplica.graphsdk"
-val uploadedBinariesVersion = "0.0.1-rc1"
+val uploadedBinariesVersion = "1.0.0"
 java.sourceCompatibility = JavaVersion.VERSION_17
 version = if (project.hasProperty("projVersion")) {
     project.properties["projVersion"]!!
 } else {
-    "0.0.2-SNAPSHOT"
+    "1.0.1-SNAPSHOT"
```
aramikm, line 20: since snapshot should be higher than the last released on incrementing the patch to make sure it would show up higher but it wouldn't matter from publishing packages perspective since we are not publishing any snapshots.

saraswatpuneet (approved):
1.0.0

PR #186, opened 2024-03-13
bugfix: rearrange target check on node

aramikm:
# Goal
The goal of this PR is address a bug that only happens on Macos with x86 architecture on node library and tries to load the windows binary since both darwin and win32 contained `win` in it.

Related to #154


aramikm:
duplicate of #185

PR #187, opened 2024-03-26
Update for Paseo support

wilwade:
# Goal
The goal of this PR is to add the Frequency Paseo Testnet support to the Graph SDK


bridge/common/protos/output.proto
```diff
@@ -38,6 +38,7 @@ enum EnvironmentType {
   MainNet = 0;
   Rococo = 1;
   Dev = 2;
+  TestnetPaseo = 3;
```
aramikm, line 41: nit: Since these protos are only used to communicate between rust and jni I think we can just add Paseo after Mainnet and shift the values

config/resources/configs/frequency-testnet-paseo.json
```diff
@@ -0,0 +1,41 @@
+{
+  "sdkMaxUsersGraphSize": 10000,
+  "sdkMaxStaleFriendshipDays": 90,
+  "maxGraphPageSizeBytes": 1024,
+  "maxPageId": 16,
+  "maxKeyPageSizeBytes": 65536,
+  "graphPublicKeySchemaId": 7,
+  "schemaMap": [
+    [
+      8,
```
aramikm, line 10: Just wanted to make sure that these schema id's are the same as what we registered on paseo for each

java/example-graphsdk-client/src/test/kotlin/io/amplica/graphsdk/LibraryTest.kts
```diff
@@ -31,6 +31,16 @@ class LibraryTest {
         graph.finalize()
     }
 
+    @Test
```
aramikm, line 34: We should remove this. This tries to test against the uploaded jar but since the uploaded jar does not have `getTestnetPaseo()` it is failing.

aramikm (approved):
Looks good, some nits and a comment to fix the ci issue

config/resources/configs/frequency-testnet-paseo.json
```diff
@@ -0,0 +1,41 @@
...
+      10,
+      [
+        "1.0",
+        {
+          "connectionType": "friendship",
+          "privacyType": "private"
+        }
+      ]
+    ]
+  ],
```
saraswatpuneet, line 39: On the paseo schema 10 does not have any settings , expecting signature required setting on it ?

aramikm:
Might want to add a test like `getGraphConfig with Rococo environment should return the graph config` for paseo in node bridge

config/resources/configs/frequency-testnet-paseo.json
saraswatpuneet: Same with schema 9 it has setting set to 0 , I think private schemas require signatures enforcement

saraswatpuneet: schema 8 is good it's Avro Paginated and setting=0

bridge/node/js/models/environment.ts
```diff
@@ -4,6 +4,7 @@ enum EnvironmentType {
   Mainnet = "Mainnet",
   Rococo = "Rococo",
   Dev = "Dev",
+  TestnetPaseo = "TestnetPaseo",
 }
 
 interface EnvironmentInterface {
```
aramikm, line 10: line 15
`environmentType: EnvironmentType.Mainnet | EnvironmentType.Rococo | EnvironmentType.TestnetPaseo;`

config/resources/configs/frequency-testnet-paseo.json
saraswatpuneet: I think schema 7 has setting = 3 not sure if it corresponds to `AppendOnly`

config/resources/configs/frequency-testnet-paseo.json
wilwade: They are. Here's the output of the deploy script:

```
Generated schema mapping:
 {
  "0x203c6838fc78ea3660a2f298a58d859519c72a5efdc0f194abd6f0d5ce1838e0": {
    "tombstone": {
      "1.2": 1
    },
    "broadcast": {
      "1.2": 2
    },
    "reply": {
      "1.2": 3
    },
    "reaction": {
      "1.1": 4
    },
    "profile": {
      "1.2": 5
    },
    "update": {
      "1.2": 6
    },
    "public-key-key-agreement": {
      "1.2": 7
    },
    "public-follows": {
      "1.2": 8
    },
    "private-follows": {
      "1.2": 9
    },
    "private-connections": {
      "1.2": 10
    },
    "public-key-assertion-method": {
      "1.3": 11
    }
  }
}
```

config/resources/configs/frequency-testnet-paseo.json
wilwade: These are the same settings as appear on mainnet. So if they are wrong, so is mainnet 😟

saraswatpuneet: Yeah I think they are wrong "in spirit" meaning most stateful operations everywhere is only calling "upsert_page" extrinsic and it does not check for schema permission but extrinsics with "...._with_signature" will fail 

And I think delete page extrinsic will fail

saraswatpuneet (approved):
lgtm! schema related questions does not affect this PR

PR #188, opened 2024-04-01
fix: when generating page_capacities.rs, do not format other source file

JoeCap08055:
# Goal
The goal of this PR is to avoid re-formatting all source files when we only want to generate a new `page_capacities.rs` file.


saraswatpuneet (approved):
Lgtm

aramikm (approved):

PR #189, opened 2024-04-01
fix: remove unnecessary borrow/clone

JoeCap08055:
# Goal
The goal of this PR is to remove some unnecessary borrowing + cloning when registering Graph update events. This should reduce some transient short-term memory usage.

### DIFF
--- a/core/src/graph/updates.rs
+++ b/core/src/graph/updates.rs
@@ -49,10 +49,10 @@ impl UpdateTracker {
 	#[log_result_err(Level::Info)]
 	pub fn register_update(
 		&mut self,
-		event: &UpdateEvent,
+		event: UpdateEvent,
 		ignore_existing: bool,
 	) -> DsnpGraphResult<()> {
-		if self.contains(event) {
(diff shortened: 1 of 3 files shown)


core/src/graph/updates.rs
```diff
@@ -83,14 +83,14 @@ impl UpdateTracker {
 	#[log_result_err(Level::Info)]
 	pub fn register_updates(
 		&mut self,
-		events: &[UpdateEvent],
+		mut events: Vec<UpdateEvent>,
```
aramikm, line 86: I don't think this `mut` is necessary.  you can just use `into_iter() ` on the vec and it will give you the item

JoeCap08055: Even if I'm using `.drain(..)` below?

JoeCap08055: fixed; changed `.drain(..)` to `.into_iter()` and removed `mut`.

saraswatpuneet (approved):

PR #191, opened 2024-04-01
Fixed existing bug on transaction hashmap

### LINKED ISSUE #190: [Bug] Out of memory bug
# Description

Having lots of connections ~7k or more causes an Out of memory issue.

# Acceptence criterea
- Fix the excessive memory usage in graph-sdk.

aramikm:
# Goal
The goal of this PR is to address the bug in `transactional-hashmap` that instead of only keeping track of original state it was keeping track of all the intermediate states which was allocating excessive memory and causing Out of memory errors.

Closes #190 

# Discussion
- Fixed the implementation of transaction-hashmap to only keep track of the persisted state

# Checklist
- [x] Tests added

### DIFF
--- a/core/src/util/transactional_hashmap.rs
+++ b/core/src/util/transactional_hashmap.rs
@@ -12,19 +12,13 @@ pub trait Transactional {
 }
 
 #[derive(Debug, Eq, PartialEq, Clone)]
-enum Reversible<K, V> {
-	Add { key: K, prev: Option<V> },
-	Remove { key: K, prev: V },
-}
-
-#[derive(Debug, Eq, PartialEq, Clone)]
 pub struct TransactionalHashMap<K, V>
 where
 	K: Eq + Hash + Clone,
 	V: Clone,
 {
 	inner: HashMap<K, V>,
-	rollback_operations: Vec<Reversible<K, V>>,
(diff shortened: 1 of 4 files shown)


core/src/util/transactional_hashmap.rs
```diff
@@ -142,18 +146,14 @@ where
...
-			match op {
-				Reversible::Add { prev, key } => match prev {
-					Some(old) => self.inner.insert(key, old),
-					None => self.inner.remove(&key),
-				},
-				Reversible::Remove { prev, key } => self.inner.insert(key, prev),
+		for (key, v) in self.original_items.drain() {
+			match v {
+				Some(old) => self.inner.insert(key, old),
+				None => self.inner.remove(&key),
```
JoeCap08055, line 156: Why not just:
```
self.inner = self.original_items;
self.original_items = HashMap::new();
```

aramikm: we can't, originals only keeps track of the originals that are overridden. Maybe I need to rename it to `overriden_originals` and inner is the latest state for all items.

saraswatpuneet (approved):
Lgtm! Joe have a better idea about fixes so consider this as soft approval

PR #193, opened 2024-04-02
Expose commit/rollback via bridge apis

### LINKED ISSUE #192: Bridge APIs do not expose commit/rollback
# Description
As an API user, I want to be able to take full advantage of GraphSDK's `commit`/`rollback` feature. Currently, this feature is only available to the native Rust API; developers using Node.js, Java, or C-compatible FFI should also be able to take advantage of this feature.

# Acceptance Criteria
- [ ] Ability to apply actions to a graph without auto-committing them
- [ ] `commit` and `rollback` methods exposed via bridge APIs

JoeCap08055:
# Goal
The goal of this PR is to expose GraphSDK's commit/rollback functionality to the Bridge APIs (currently only available via native Rust API)

* Add `disable_auto_commit` flag to `ActionOptions`, have `apply_actions` honor it
* Expose commit/rollback to Node.js, Java, and C APIs

Closes #192 

# Checklist
- [x] Tests added


JoeCap08055:
I'll sync this from main once #191 is merged

.cargo-deny.toml
```diff
@@ -150,7 +150,7 @@ exceptions = [
 # The name of the crate the clarification applies to
 #name = "ring"
 # The optional version constraint for the crate
-version = 2
+#version = 2
```
aramikm, line 153: nit: was there any CI issue with this?

bridge/node/src/api.rs
```diff
@@ -391,6 +392,56 @@ pub fn apply_actions(mut cx: FunctionContext) -> JsResult<JsBoolean> {
...
+pub fn commit(mut cx: FunctionContext) -> JsResult<JsBoolean> {
+	let graph_state_id: Handle<'_, JsNumber> = cx.argument::<JsNumber>(0)?;
+	let graph_state_id = graph_state_id.value(&mut cx) as usize;
+
+	let mut states = GRAPH_STATES.lock().unwrap();
+	let graph_state = states.get_mut(&graph_state_id);
+	if graph_state.is_none() {
+		return cx.throw_error("Graph state not found");
+	}
+	let graph_state = graph_state.unwrap();
```
aramikm, line 412: question: I've noticed we have a lot of unwraps in this file. Do these gets catches somewhere else or it might cause unexpected behavior if it happens?

aramikm (approved):
Looks good. Thanks for adding this. A couple of nits and I think we might also need a small test on java side to ensure jni works

.cargo-deny.toml
JoeCap08055: Yes, was failing CI

JoeCap08055: Or maybe not CI, but failed when I did `make all` locally

bridge/node/src/api.rs
saraswatpuneet: Same for all the places where unwrap is added

PR #195, opened 2024-04-03
bugfix: fixed next page id calculation

### LINKED ISSUE #194: [Bug]
# Description

When importing a lot of users on a non-existing graph the page calculation always returns 0 since it does not count the updated pages. Which in result only exports 1 page instead of all of them.

# Acceptence Criterea
- Use updated pages in next_page_id calculations

aramikm:
# Goal
The goal of this PR is fix the bug which exports only the last page instead of all of them when more than 1 page is created via adding connections.

Closes #194 

# Checklist
- [X] updated versions
- [X] Tests added

### DIFF
--- a/core/src/api/api.rs
+++ b/core/src/api/api.rs
@@ -733,7 +733,7 @@ mod test {
 	}
 
 	#[test]
-	#[timeout(30000)]
+	#[timeout(100000)]
 	fn add_large_number_of_follows_to_private_follow_graph_should_succeed() {
 		// arrange
 		let env = Environment::Mainnet;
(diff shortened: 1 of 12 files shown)


core/src/graph/graph.rs
```diff
@@ -890,7 +899,35 @@ mod test {
 			))),
 		};
 
-		assert_eq!(graph.get_next_available_page_id(), Some(8));
+		assert_eq!(graph.get_next_available_page_id(&BTreeMap::default()), Some(8));
+	}
+
+	#[test]
+	fn get_next_available_page_should_include_updated_pages() {
```
JoeCap08055, line 906: 💯

JoeCap08055 (approved):

core/src/api/api.rs
```diff
@@ -733,7 +733,7 @@ mod test {
 	}
 
 	#[test]
-	#[timeout(3000)]
+	#[timeout(30000)]
```
JoeCap08055: Timed out in CI
```suggestion
	#[timeout(60000)]
```

saraswatpuneet (approved):

PR #196, opened 2024-05-01
Graph simulator init

aramikm:
# Goal
The goal of this PR is to allow us to do some large scale verification and stress testing.

Closes 

# Discussion

# Checklist
- [ ] Tests added


simulator/README.md
```diff
@@ -0,0 +1,19 @@
+# Graph Simulator
+
+This is a stand-alone utility that allows us to run different scenarios on the whole of social graph and test
+the desired behavior against the sdk latest implementation.
+
+## Initialized State
+Calculating the initial state for all te graphs and keys is time-consuming, and we've already calculated
```
JoeCap08055, line 7:
```suggestion
Calculating the initial state for all the graphs and keys is time-consuming, and we've already calculated
```

simulator/src/scenarios/key_rotation.rs
```diff
@@ -0,0 +1,78 @@
...
+		PRIVATE_FRIENDSHIP_PAGE_MODIFICATIONS,
+	},
+	scenarios::common::{add_keys_for_users, compare_on_chain_with_expected, modify_random_pages},
+	GlobalState,
+};
+use dsnp_graph_config::Environment;
+use dsnp_graph_core::api::api_types::*;
+use rand::{prelude::SliceRandom, thread_rng};
+
+pub fn execute_key_rotation_private_follow(state: &mut GlobalState, env: Environment) {
```
JoeCap08055, line 13: Hmm, this makes me think... I know we talked about a future enhancement to support a key rotation top-level method in the API, but I don't think we ever created an issue for it...

simulator/src/scenarios/common.rs
```diff
@@ -0,0 +1,204 @@
...
+						owner_dsnp_user_id: *user_id,
+						new_public_key: graph_key_pair.public_key.clone(),
+					}],
+					&None,
+				)
+				.expect("error adding graph key");
+
+			let updates = graph.export_updates().expect("error exporting updates");
+
+			println!("importing keys for user {}", user_id);
```
JoeCap08055, line 54: This is REALLY noisy... (also in one other place, same message). Maybe we don't need this debug line?

JoeCap08055 (commented):
When running without the 20000 users pre-loaded data, I see lots of messages about ```
thread '<unnamed>' panicked at simulator/src/init.rs:145:18:
error exporting updates with 6099 connections: GraphIsFull
```

When running with the 20000 pre-load file, I see the same errors, but the process finishes pretty quickly.

Both runs end with the same result:
```
make: *** [run-sim] Error 101
```

File: simulator/README.md (hunk 1 above)
Comment (saraswatpuneet) [reply simulator/README.md:7]: nit: Also we should mention to build the sdk first, my ran picked up old binary and failed

**Review by saraswatpuneet — APPROVED** [review]
left some nit
* ran fine, i had to update maxPageId in all 3 configs but not sure which is right one
* we can update the doc to reflect the above

**Comment by JoeCap08055** [comment]
> When running without the 20000 users pre-loaded data, I see lots of messages about
> 
> ```
> thread '<unnamed>' panicked at simulator/src/init.rs:145:18:
> error exporting updates with 6099 connections: GraphIsFull
> ```
> 
> When running with the 20000 pre-load file, I see the same errors, but the process finishes pretty quickly.
> 
> Both runs end with the same result:
> 
> ```
> make: *** [run-sim] Error 101
> ```
> 
> Steps taken:
> 
> * Edit _frequency.json_ to change max page id to 64
> * Put `20000_simulation_state.bin` in project root
> * Run `make build-sim ; make run-sim`

Found my error; re: @saraswatpuneet 's comment about building the SDK first. I had a stale build from an older branch.
Simulation runs successfully now.

File: simulator/src/scenarios/common.rs (hunk 3 above)
Comment (JoeCap08055) [reply simulator/src/scenarios/common.rs:54]: Hmm, not so noisy now that I am running with the right version of the SDK built 😆

File: simulator/README.md [hunk 4]
```diff
@@ -0,0 +1,19 @@
...
+the state for 20000 users in the graph which is uploaded [here](https://drive.google.com/file/d/1WoQ_INPa4aOy4dCm1WUN9KXxq08P0-Rp/view?usp=sharing)
+and if copied inside repository's root folder it can skip the initialization process, and you can directly run
+test scenarios on already established state.
+
+### Build
+For being able to run the 20k users example the **maxPageId** on _frequency.json_ file should be changed to at least **64**.
+We will update this value after it was updated un Frequency chain first.
```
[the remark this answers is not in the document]
aramikm: based on what is mentioned in the Readme you should only update `frequency.json`. I will add that you will need ot build after so that it rans the latest version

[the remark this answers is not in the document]
aramikm: @JoeCap08055 If we would like to test the any private friendship scenarios we need at least 64 but since we don't have any use-cases on-chain for it it doesn't make sense to set the chain value to 64 for now.

JoeCap08055 (approved):
* Built SDK and ran simulation
lgtm

simulator/src/scenarios/key_rotation.rs
aramikm: We have `force_recalculate_graphs` that does key rotation on all pages also but currently we only do key rotation lazily whenever a page is modified.

